Issue #20, opened 2019-09-09
Docstrings are missing from many files


stevetorr:
Docstrings annotating arguments and outputs to various methods would improve user readability.

Issue #16, opened 2019-09-05
Serialization and representation (e.g. __str__) methods for various objects


stevetorr:
Methods for serializing certain objects which are passed between models (i.e. atomic environments, structures, etc), or even models themselves, would be useful. The advantage of this over pickled objects is that they can be more human-readable (and I understand that pickled objects have some security risks associated with them). 

One example application is that JSON objects are easily storable in certain database architectures. This might be relevant for e.g. FOOGA in the near future if we want to automate the process of training GP models for different datasets, as this would let us store them more easily.

In my development branch I've done this for the AtomicEnvironment object. There are ways we could standardize this or easily implement it across our codebase (e.g. by using Monty, which has an object type which allows for effortless JSON serialization of different Python objects).

jonpvandermause:
Good idea to serialize, would be nice to be able to print out attributes of the env/struc/gp objects. As for saving and storing entire gp models, pickle.dump(gp_model, gp_file) works well in practice.

stevetorr:
Todo: Look through classes in env, struc, gp, and add representations

stevetorr:
We have solved serialization.

Issue #15, opened 2019-09-05
Underway: Developing new GP from AIMD toolkit


stevetorr:
I'm making this for my own development process

Issue #14, opened 2019-09-05
Predict methods in OTF should be moved outside of class


stevetorr:
Hat tip to Lixin and Yu who discovered the following problem (1) after laborious debugging:

1.Apparently parallelization for python fails when multiple processes are operating on the same instance from the same class object.

2. This also helps to keep the OTF class itself smaller and focused, while freeing up the predict functions to be used for other purposes. 

For instance, the module I'm developing of gp_from_aimd has great cause to use the predict functions, and to avoid duplicating code, having them be in a different file allows them to be called without an OTF instance. I currently have implemented this in my development branch, for reviving gp_from_aimd. Lixin has done the same in hers, so one of us will push it eventually.

jonpvandermause:
@YuuuuXie @nw13slx Is there still a parallelization bug? I remember noticing an issue with the parallelization in otf a few months ago, but patched it with 5dd377a and haven't noticed any problems since. Seems that the parallel methods as implemented work fine and give a significant speed up.

But yes, the predict on structure methods can be moved, perhaps changed to methods of the gp class.

stevetorr:
To do: Test an OTF job on Odyssey to ensure this works in parallel. Test if moving it to another file breaks anything (moving it outside of a class seems to make it work okay, says Yu).

PR #27, opened 2019-09-11
GP from AIMD + Serialization

stevetorr:
I spoke with @jonpvandermause  about grandfathering in this commit since I started working on it at the same time as when we resolved to do branch-based development and the commits are a little 'entangled' between changes. In the future I'll split up the individual features on a branched basis, but as is, this PR addresses:

Completely:
- Issue 15: https://github.com/mir-group/flare/issues/15
- Issue 14: https://github.com/mir-group/flare/issues/14  (Though the predict methods are still in OTF.py, we can delete them and refactor OTF later, as they are currently only unit tested in a limited way in OTF and on their own).

Partially:
-  Issue 20 https://github.com/mir-group/flare/issues/20
- Issue 16: https://github.com/mir-group/flare/issues/16

In summary, this PR has:

- A first version of the GP to AIMD module (with 95% unit test coverage!)
- Serialization methods for env and struc
- A JSON encoder to assist with the above serialization methods
- Added/ updated docstrings for many different classes
- PEP8 Compliance (@jonpvandermause , our editors disagree about over / under indentation but everything else should be good)

Feedback welcome:

@nw13slx  Can you weigh in on what the issue was with the predicting while in parallel? I know that you mentioned as much during the meeting last week but I forget what the exact cause was; and, if so, if the current way of splitting it in predict.py  would fix your issues.


flare/gp.py
```diff
@@ -228,6 +264,13 @@ def en_kern_vec(self, x: AtomicEnvironment) -> np.ndarray:
         return k_v
 
     def set_L_alpha(self):
+        """
+        Invert the covariance matrix, setting L (a lower triangular
+        matrix s.t. L L^T = (K + sig_n^2 I) ) and alpha, an inversion of the
+        vector of training labels.
```
jonpvandermause: Alpha is the inverse of the covariance matrix multiplied by the vector of training labels.

flare/gp_from_aimd.py
```diff
@@ -0,0 +1,232 @@
...
+                            dt=0,
+                            Nsteps=len(self.frames),
+                            structure=self.frames[0],
+                            std_tolerance=(self.rel_std_tolerance,
+                                           self.abs_std_tolerance),
+                            output_name=self.output_name)
+
+        self.start_time = time.time()
+
+    def run(self):
```
jonpvandermause: Down the line we'll want to add a number of options for training on a fixed AIMD set; might make sense to generalize the code a little now to make way for future implementations. Passing over each frame and adding environments with high uncertainty makes sense and is data efficient, but it's not the only way (or the easiest). We should also make it easy to train on equally spaced frames, user specified frames, etc. @nw13slx has been playing with different approaches --- it would be good to coordinate with her on this.

flare/gp_from_aimd.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def train_gp(self):
+        self.gp.train(monitor=True if self.verbose >= 2 else False)
+
+        output.write_hyps(self.gp.hyp_labels, self.gp.hyps,
+                          self.start_time, self.output_name,
+                          self.gp.like, self.gp.like_grad)
+        self.train_count += 1
+
+    def is_std_in_bound(self, frame):
```
jonpvandermause: Looks similar to the is_std_in_bound method of otf. To avoid code duplication, we should have it as a function in one place (either otf or a separate file) and import from there.

flare/predict.py
```diff
@@ -0,0 +1,105 @@
...
+Helper functions which obtain forces and energies
+corresponding to atoms in structures
+"""
+from flare.env import AtomicEnvironment
+from flare.gp import GaussianProcess
+from flare.struc import Structure
+
+import numpy as np
+import concurrent.futures
+
```
jonpvandermause: These methods are similar to the ones in otf.py, so this file duplicates a lot of code. Before modifying otf.py to call these functions, we'll need to check that they work properly on 32 core jobs (parallelization has been finicky in the past)

flare/gp.py
```diff
@@ -72,6 +74,28 @@ def update_db(self, struc: Structure, forces: list,
         # create numpy array of training labels
         self.training_labels_np = self.force_list_to_np(self.training_labels)
 
+        if self.l_mat is None:
+            self.set_L_alpha()
+        else:
+            self.update_L_alpha()
```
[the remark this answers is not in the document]
stevetorr: Oh, didn't know update wasn't unit tested. It now will use only set_L_alpha().

flare/gp.py
stevetorr: Whoops, thanks. Fixed!

flare/gp_from_aimd.py
stevetorr: These points are all totally valid. 

This comment touches on the tradeoff to be made here between how much power to cede to the user and how much to build in as functionality, which then results in a more complex training manager. I am pretty conflicted about this myself, and I'm not sure where the line should be. I would love to solicit input from you + @nw13slx + everyone else at a developer meeting about this.

It feels simultaneously like a huge cop out and not a cop out to say that the user can do some of this already. 

For example, if the typehint is followed, one can pass in a list of Structures sliced by e.g.
`input_list[::10]` (though that will not check the MAE against every frame in the trajectory, which might be desirable). An earlier version of the TrajectoryTrainer I had allowed 'seed structures' to be passed in, which would then be applied to the GP before the run through the entire trajectory began. I decided against leaving it in, hoping that the ability to add environments one-by-one into the GP added via the add_one_env method would enable the user to do this before putting their GP in.

On the other perspective, in which the above argument is a cop out, the whole point of TrajectoryTrainer is to automate labor and pre-processing. Musing about where 'the line' is, I think the features you request are probably on the side of inclusion. We can add a 'skip through' option and some version of 'seed frames' which can be passed to the TrajectoryTrainer. One can also do the frames pre-processing or GP pre-processing if they like, and skip the above two features anyway.

flare/gp_from_aimd.py
stevetorr: Agreed. I can transfer this to a different file and re-factor OTF to point there (note that this implementation borrows the good idea of @nw13slx to allow a relative and absolute cutoff simultaneously, and I unit-tested every permutation). I think a good home could be within  predict.py, as the functionality it serves is to be a diagnostic to a prediction made on a given atom.

flare/predict.py
stevetorr: Sounds good and I agree. We can consult @nw13slx  and @YuuuuXie ; I know at least Lixin has worked with an implementation similar to this on her branch.

Issue #36, opened 2019-09-12
output file shouldn't be repeatedly open and close


nw13slx:
The output file in output.py should not be repeatedly open and close. And we also need to allow multiple output files

Issue #37, opened 2019-09-12
Hyperparameter output to a separate file


stevetorr:
Would help to make managing the results of runs easier.

PR #42, opened 2019-09-13
new output class and add configuration output in ".xyz" format

nw13slx:
A new class Output is set to replace the original output module. The class opens a log file and two xyz files at the init function and closes all files in conclude_run() function. All the original functions in the output module now become function members in the Output class, with the same arguments.
new class for output and write atomic configurations to xyz files
The new class avoid open/close the log file too often and also automatically back up the old log file if it exists before the current run starts.

write_xyz_config(...) function is also added to print configs in xyz format. It is called in the record_state function in otf.py and md_run.py.

Completely:
- issue #36

### DIFF
--- a/flare/gp.py
+++ b/flare/gp.py
@@ -24,7 +24,7 @@ class GaussianProcess:
                  energy_kernel: Callable = None,
                  opt_algorithm: str = 'L-BFGS-B',
                  maxiter=10, par=False,
-                 monitor=False):
+                 output=None):
         """Initialize GP parameters and training data."""
 
         self.kernel = kernel
@@ -45,7 +45,7 @@ class GaussianProcess:
         self.likelihood = None
         self.likelihood_gradient = None
         self.par = par
-        self.monitor = monitor
+        self.output = output
 
     # TODO unit test custom range
     def update_db(self, struc: Structure, forces: list,
@@ -112,7 +112,7 @@ class GaussianProcess:
 
         return forces_np
 
-    def train(self, monitor=False, custom_bounds=None,
+    def train(self, output=None, custom_bounds=None,
               grad_tol: float = 1e-4,
               x_tol: float = 1e-5,
               line_steps: int = 20):
@@ -126,7 +126,7 @@ class GaussianProcess:
         x_0 = self.hyps
 
         args = (self.training_data, self.training_labels_np,
-                self.kernel_grad, self.cutoffs, monitor,
+                self.kernel_grad, self.cutoffs, output,
                 self.par)
 
         if self.algo == 'L-BFGS-B':
--- a/flare/gp_algebra.py
+++ b/flare/gp_algebra.py
@@ -261,11 +261,11 @@ def get_like_grad_from_mats(ky_mat, hyp_mat, training_labels_np):
 
 def get_neg_likelihood(hyps: np.ndarray, training_data: list,
                        training_labels_np: np.ndarray,
-                       kernel, cutoffs=None, monitor: bool = False,
+                       kernel, cutoffs=None, output = None,
                        par=False):
 
-    if monitor:
-        print('hyps: ' + str(hyps))
+    if output is not None:
+        output.write_to_log('hyps: ' + str(hyps), name="hyps")
 
     if par:
         ky_mat = \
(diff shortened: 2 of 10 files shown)


nw13slx:
@jonpvandermause   all test ran well on my python environment on Odyssey... could you please show the syntax error? so I can figure out what's wrong?

jonpvandermause:
Turns out I was 12 commits behind. This was the offending line (after commit 20374f9, since been corrected):
image

The tests work at the tip of the branch, but after running them the following files were generated and flagged by git after "git status":
image

Should we update the .gitignore so that these aren't tracked?

nw13slx:
The latest commit should also solve issue #37 . All the hyper-parameter optimization output is redirected to a file with "-hyps.dat" suffix.

flare/qe_util.py
```diff
@@ -9,7 +9,7 @@
 def run_espresso(qe_input, structure, pw_loc):
     run_qe_path = qe_input
     edit_qe_input_positions(run_qe_path, structure)
-    qe_command = 'mpirun {0} < {1} > {2}'.format(pw_loc, run_qe_path,
+    qe_command = '{0} < {1} > {2}'.format(pw_loc, run_qe_path,
```
stevetorr: This is a good catch, puts more power in the hand of the user to define the PWSCF command with/without MPI

flare/qe_util.py
```diff
@@ -24,6 +24,8 @@ def run_espresso_par(qe_input, structure, pw_loc, no_cpus):
         'mpirun -np {0} {1} < {2} > {3}'.format(no_cpus, pw_loc, run_qe_path,
                                                 'pwscf.out')
 
+    with open("alog", "a+") as fout:
```
stevetorr: What does alog mean / stand for?

flare/gp_algebra.py
```diff
@@ -278,20 +278,19 @@ def get_neg_likelihood(hyps: np.ndarray, training_data: list,
 
     like = get_like_from_ky_mat(ky_mat, training_labels_np)
 
-    if monitor:
-        print('like: ' + str(like))
-        print('\n')
+    if output is not None:
+        output.write_to_log('like: ' + str(like)+'\n', name="hyps")
```
stevetorr: A suggestion which is somewhat to personal preference: I personally find formatting strings using
` 'like: {} \n'.format(like) `
to be more readable.
I know that you also just translated the existing line this way, but going forward we could do all of our string formatting this way.

flare/gp_from_aimd.py
```diff
@@ -244,11 +241,11 @@ def train_gp(self):
         """
         Train the Gaussian process and write the results to the output file.
         """
-        self.gp.train(monitor=True if self.verbose >= 2 else False)
+        self.gp.train(output=self.output if self.verbose >= 2 else None)
```
stevetorr: I like the way you implemented this.

flare/gp.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, kernel: Callable,
                  energy_kernel: Callable = None,
                  opt_algorithm: str = 'L-BFGS-B',
                  maxiter=10, par=False,
-                 monitor=False):
+                 output=None):
```
stevetorr: Here and in a few other places, we could typehint output as the `Output` object.

flare/output.py
```diff
@@ -2,204 +2,320 @@
...
+        else:
+            self.outfiles[filetype] = open(filename, "w+")
+
+    def write_to_log(self, logstring: str, name: str = "log"):
+        """
+        Write any string to logfile
+        """
+        self.outfiles[name].write(logstring)
+
+    def write_header(self, cutoffs, kernel_name,
```
stevetorr: In the future, we should consider writing a more general function that takes a dictionary in as an argument and then prints the key, value pairs in a similar way that the header method does (as well as a few other methods in this document which are tailored to individual use cases).

flare/gp.py
jonpvandermause: Type hints will be eliminated eventually, as they don't play well with the Sphinx documentation. Might be better to do without them.

stevetorr: Sad to hear that Sphinx can't handle them. That's a shame. At least docstrings can contain the same info.

jonpvandermause: Actually it looks like Sphinx only really struggles with np.ndarray type hints. Here's an example from the docs:
image

If we can find a way to get Sphinx to format that better, type hints should be okay.

flare/qe_util.py
nw13slx: ah... It's a line I left for debug. I forgot to delete it...

PR #46, opened 2019-09-13
fix import in mff test

### LINKED ISSUE #45: mc kernel issue for test_mff.py
@YuuuuXie , I think that the import statement needs to be reformatted in the test_mff.py file, as it references the mc_kernels module which has since moved / changed:

image

It results in an error seen above.

jonpvandermause:
This should close #45 @stevetorr


jonpvandermause:
Yes, it requires an environment variable $lmp pointing to a working lammps executable, and also requires that the mff pairstyle is installed. Makes sense to skip it if there is no lmp variable.

Issue #47, opened 2019-09-13
Unit tests should skip instead of fail for users without QE


stevetorr:
As is, the tests which involve calling Espresso fail if the PWSCF command is not found. 
It would be nice if the PWSCF command was not detected for the unit tests for calling QE to not fail, but simply be skipped. This may be more informative to the user, as there are reasons why it could fail in attempting to call QE which would require debugging. 
We could also print a message encouraging the user to fix their environment variable.

stevetorr:
Working on this now, can be done with the following command:

@pytest.mark.skipIf(not os.environ.get('PWSCF_COMMAND',
                          False), 'PWSCF_COMMAND not found '
                                  'in environment: Please install Quantum '
                                  'ESPRESSO and set the PWSCF_COMMAND env. '
                                  'variable to point to pw.x.')

PR #48, opened 2019-09-13
Now skips tests with QE if PWSCF_COMMAND is not found

stevetorr:
Now skips unit tests which have no hope of succeeding if the user does not have the PWSCF_COMMAND environment variable:

image

Also tightened up an import statement in test_qe_util. Didn't think it warranted an issue.


stevetorr:
Closes #47

Issue #29, opened 2019-09-11
"mff" should no longer be used, the name is gonna change


YuuuuXie:
to "mgp" (Mapped Gaussian Process) (credit to Jon)

Issue #23, opened 2019-09-09
update_L_alpha and update_L_alpha_v1 in gp.py need unit tests


jonpvandermause:
The tests should check that the resulting matrices agree with set_L_alpha.

PR #50, opened 2019-09-13
changed mff name #29 & tested update_L_alpha #23

YuuuuXie:
1. "mff" modules have been changed to "mgp", code is cleaned and tested by `test_mgp.py`
2. `update_L_alpha` is tested in `test_gp.py`


jonpvandermause:
@YuuuuXie Looks good. Should we delete the mff folder?

jonpvandermause:
test_mgp failed for me:
image

Here are the line that triggered the error:
image

Looks like this should be "MappedGaussianProcess" instead of "MappedForceField".

YuuuuXie:
@jonpvandermause I just fixed it. In my local repo, I've things in mff folder to mgp folder, so I think you can remove the mff folder in the remote repo

jonpvandermause:
Still failing for me, this time because subprocess is used in mgp.py but not imported at the top:
image

YuuuuXie:
This is weird because it works for me
Screen Shot 2019-09-14 at 2 41 18 PM

jonpvandermause:
Ah, based on your screenshot it looks like the test is getting skipped because you don’t have $lmp as an environment variable pointing to a lammps executable. Could you try adding that and rerunning the test?

YuuuuXie:
Yes you're right. I made a new "push"

jonpvandermause:
Still failing, this time because the variable 'a12' isn't defined on line 515 of mgp.py. Here's the failure message:
image

Here's the part of mgp.py that triggered it, looks like '_' should be replaced with 'a12':
image

The corresponding part of mff.py defined a12 here:
image

YuuuuXie:
I noticed that I didn't succeed pushing my local repo's mgp.py to the remote repo, though it has already been modified & passed the unit test. And just now I saw that I get behind the "master" branch so I made a "git pull origin master" now my modified file is uploaded.

One question about git, if I didn't "git pull origin master", does it mean that some commit will not be made?

jonpvandermause:
All tests are working now, thanks! I just deleted the mff folder and test_mff.py, so we've officially transitioned to mgp.

Pulling shouldn't be needed to update the remote -- it's okay that the branch got behind master, because once we merge the pull request, they'll be even again. But it's possible that you had some commits on your local branch that weren't pushed. Commits don't automatically update the remote, so after a series of commits you have to run "git push" for the remote branch to get updated. That may have been the issue, but I'm not certain...

Issue #19, opened 2019-09-09
VASP parsers and utility


stevetorr:
Would be helpful to (short term) provide built-in methods to parse VASP files for model training, and (long term) support VASP interface for OTF runs

YuuuuXie:
One possible solution is to use the ASE interface, which supports various types of calculators. I'll try updating the ASE interface by this week's meeting

PR #61, opened 2019-09-20
create dft_interface module to interface with multiple DFT software.

### LINKED ISSUE #14: Predict methods in OTF should be moved outside of class

### LINKED ISSUE #56: flexible interface for DFT call
We need to reform the OTF class to interface with

- QE
- CP2K
- ASE
- Pymatgen (VASP)

nw13slx:
a new modeul "dft_interface" is set up in to allow a choice of software in OTF training.

1. user can specify DFT software by setting dft_softwarename="qe", or "cp2k" in the OTF init function. 
2. generic interface available for any future DFT interfaces
3. all function names in qe_util.py are changed to be more generic. Instead of "xx_pw_" and "xx_espresso_", the functions are all named as "xx_dft_"
3. all predict functions in OTF.py is removed. now they only rely on the functions in predict.py
4. the function argument for predict_on_atom is shrink to one tuple argument. so that it can be executed with multiple threads
5. new unit tests for parallel OTF run with QE and CP2K are added.

closes #56 closes #14 

This interface should make it much easier to have pymatgen and ase parser as mentioned in #19


nw13slx:
@anjohan   Could you please help checking out how to install CP2K-6 in the Travis virtual machine properly? The Ubuntu16 LST repo only has CP2K-2...

Issue #63, opened 2019-09-22
Multiple versions of `like` and `likelihood` variables floating around in gp.py


stevetorr:
There is some redundancy in certain methods setting the `like` or `likelihood` variable in different places.

PR #64, opened 2019-09-22
Serialization and representation GP + Small fixes

stevetorr:
Introduces a serialization method for GPs (to_ and from_dict methods) and a representation (__str__) method. Introduced a new string-to-kernel method in kernels.py to facilitate the from_dict method. The serialization is unit tested. Fixed a few style errors in other parts of the code. A method in gp.py sets a 'self.like' attribute and others set `self.likelihood`. This is also now corrected towards `self.likelihood`.


stevetorr:
Closes #63 , addresses #16

dmclark17:
Regarding the string to kernel logic, you may be able to refactor the function to use `getattr` and `getattr`. You might be able to get away with not needing to define a dictionary mapping strings to functions. 

Unclear if this would be better though 🤷‍♂

stevetorr:
> `numpy.ndarray`s are not JSON serializable using the default encoder. Is there a canonical way to store them on disk for the project?

I've been using an encoder that I found online which is included in util.py to handle converting numpy arrays to JSON.


> Regarding the string to kernel logic, you may be able to refactor the function to use `getattr` and `getattr`. You might be able to get away with not needing to define a dictionary mapping strings to functions.

Thanks for looking closely at the methods! What would the `getattr` correspond to- getattr called on the file itself (to refer to individual callables within the flare.kernels file)? Also can you clarify what the second `getattr` was supposed to be?

dmclark17:
I see, I didn't see the `NumpyEncoder` 👍 


Whoops that was suppose to be `getattr` and `hasattr`. You can call them on the module itself by doing something like:

```python
current_module = sys.modules[__name__]  # Gets the current module
if hasattr(current_module, "two_body"):
    func = getattr(current_module, "two_body")
```

flare/gp.py
```diff
@@ -185,12 +195,6 @@ def predict(self, x_t: AtomicEnvironment, d: int) -> [float, float]:
         pred_var = self_kern - \
                    np.matmul(np.matmul(k_v, self.ky_mat_inv), k_v)
 
-        # # get predictive variance (possibly slow)
-        # v_vec = solve_triangular(self.l_mat, k_v, lower=True)
-        # self_kern = self.kernel(x_t, x_t, self.bodies, d, d, self.hyps,
-        #                         self.cutoffs)
-        # pred_var = self_kern - np.matmul(v_vec, v_vec)
-
```
jonpvandermause: Nice, these lines needed to go.

flare/gp.py
```diff
@@ -292,8 +296,8 @@ def set_L_alpha(self):
         self.ky_mat_inv = ky_mat_inv
         self.l_mat_inv = l_mat_inv
 
-        self.like = like
-        self.like_grad = like_grad
+        self.likelihood = like
+        self.likelihood_gradient = like_grad
```
jonpvandermause: Good catch, not sure why there were multiple versions of the likelihood.

flare/gp.py
```diff
@@ -331,3 +335,82 @@ def update_L_alpha(self):
...
+        thestr += 'Hyperparameters: \n'
+        if self.hyp_labels is None:
+            # Put unlabeled hyperparameters on one line
+            thestr = thestr[:-1]
+            thestr += str(self.hyps) + '\n'
+        else:
+            for hyp, label in zip(self.hyps, self.hyp_labels):
+                thestr += "{}: {}\n".format(label, hyp)
+
+        return thestr
```
jonpvandermause: Slick.

stevetorr:
Thanks for taking the time to look through and provide feedback!

PR #67, opened 2019-09-24
Added to / from PMG methods, with pymatgen as an optional import

stevetorr:
Talked to @dmclark17 about how to handle optional import statements for different libraries. The solution that is apparently the consensus on StackOverflow is to do as I have done below. Would be happy for 2 cents on if this makes sense.

### DIFF
--- a/flare/struc.py
+++ b/flare/struc.py
@@ -204,7 +204,8 @@ cell vectors.
     @staticmethod
     def from_pmg_structure(structure):
         """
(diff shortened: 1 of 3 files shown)


jonpvandermause (approved):
Looks good!

dmclark17:
Since we added pymatgen to the requirements.txt, we could probably just make it mandatory?

PR #69, opened 2019-09-24
Added flexibility to user to frequency of output writing

stevetorr:
I have been iterating with train_gp_from_AIMD and have found that the Output module only flushing its buffer periodically and therefore intermittently writing to the output files, while computationally more efficient in general, is undesirable for watching it run in real-time. I propose adding a few arguments to the Output class which make it write instantly to the output if the user wishes to follow output in real time and otherwise does not change the functionality of the module.


jonpvandermause:
Could we do the same for otf.py?

PR #71, opened 2019-09-24
Bug/jon/gp update

### LINKED ISSUE #70: small bug in update_db method of gp
#27 snuck a call to set_L_alpha() into the update_db method of gp. This leads to inefficiencies when constructing large gp models based on hundreds of structures, since it requires the covariance matrix to be updated from scratch every time a new structure is added to the training set. Eliminating the call to set_L_alpha resolves the issue.

jonpvandermause:
Closes #70 
Also makes gp.par=True the default in the otf parser, making it more efficient to reconstruct large gp models from otf output files.

### DIFF
--- a/flare/gp_from_aimd.py
+++ b/flare/gp_from_aimd.py
@@ -156,6 +156,7 @@ class TrajectoryTrainer(object):
                     train_atoms.append(atom)
 
             self.update_gp_and_print(frame, train_atoms, train=False)
+        self.gp.set_L_alpha()
 
         # These conditions correspond to if either the GP was never trained
         # or if data was added to it during the pre-run.
@@ -207,6 +208,8 @@ class TrajectoryTrainer(object):
 
                 if self.train_count < self.max_trains:
                     self.train_gp()
+                else:
+                    self.gp.set_L_alpha()
 
         self.output.conclude_run()
 
@@ -234,7 +237,6 @@ class TrajectoryTrainer(object):
 
         # update gp model
         self.gp.update_db(frame, frame.forces, custom_range=train_atoms)
-        self.gp.set_L_alpha()
 
         if train:
             self.train_gp()
--- a/flare/gp.py
+++ b/flare/gp.py
@@ -174,6 +174,11 @@ class GaussianProcess:
         # Kernel vector allows for evaluation of At. Env.
         k_v = self.get_kernel_vector(x_t, d)
 
+        # Guarantee that alpha is up to date with training set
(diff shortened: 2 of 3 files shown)


stevetorr:
Please forgive + correct me if this is wrong: But does this not create the danger that the training set is updated, but the `l_mat` and `alpha` attributes of the GP are not updated (where `alpha` is later used for prediction)?

stevetorr:
The alternative would be to push the responsibility onto e.g. an OTF or GP from AIMD training process, or we could include a quick check in the `predict` method to see if the length of the training set is appropriate for the size of the `alpha`, and if not, to then call `set_L_alpha` (or, if it is working, `update_L_alpha`)

jonpvandermause:
That would only happen if the gp object is used incorrectly. update_db is for updating the training database, set_L_alpha and update_L_alpha are for updating the covariance matrix and related vectors.

jonpvandermause:
Right, responsibility should be on the user, it's not worth a several order of magnitude reduction in speed for training large GPs.

jonpvandermause:
BTW, gp_from_aimd, otf, and otf_parser all call set_L_alpha immediately after update_db, so this fix also eliminates redundant calls to set_L_alpha.

flare/gp_from_aimd.py
```diff
@@ -156,6 +156,7 @@ def pre_run(self):
                     train_atoms.append(atom)
 
             self.update_gp_and_print(frame, train_atoms, train=False)
+        self.gp.set_L_alpha()
```
jonpvandermause: Good catch, this is a lot more efficient.

stevetorr:
> That would only happen if the gp object is used incorrectly. update_db is for updating the training database, set_L_alpha and update_L_alpha are for updating the covariance matrix and related vectors.

I understand what the methods do on their own, I am only pointing out the risk that users who are less familiar with the code might not realize that the predict method will not immediately work as they think it will if they are working outside of OTF, GP_from_AIMD, or OTF_Parser.

> Right, responsibility should be on the user, it's not worth a several order of magnitude reduction in speed for training large GPs.

To be clear, I don't disagree with the fix you made. I am clarifying that I'm understanding the implications for potential user confusion/unexpected behavior correctly. 

Also, length calls for lists or sizes of numpy arrays are made in constant time, so adding a line to predict to double-check that alpha is the same length would add a negligible amount to the runtime compared to e.g. the get_kernel_vector method.

stevetorr:
Feel free to disagree with the constant time check; either way, I approve of this PR!

PR #76, opened 2019-09-26
Feature/jon/npool

### LINKED ISSUE #73: adding npool option to QE otf jobs
It would be helpful to have the option to parallelize efficiently over multiple compute nodes (i.e. with the -npool flag) for large, expensive otf simulations. Should be an easy fix -- just have to give the option in otf.py to use "run_dft_npool" instead of "run_dft_par".

jonpvandermause:
closes #73 

@nw13slx I added a dummy "npool" input to the cp2k run_dft_par function so that it's still compatible with otf.py. Not sure how cp2k handles parallelization by k-point, though (or if it even can).

### DIFF
--- a/flare/dft_interface/qe_util.py
+++ b/flare/dft_interface/qe_util.py
@@ -29,7 +29,8 @@ def run_dft_par(qe_input, structure, dft_loc, no_cpus=1, dft_out='pwscf.out',
                                              dft_out)
 
     if (no_cpus > 1):
-        dft_command = 'mpirun -np {} {}'.format(no_cpus, dft_command)
+        # dft_command = 'mpirun -np {} {}'.format(no_cpus, dft_command)
+        dft_command = 'srun -n {} --mpi=pmi2 {}'.format(no_cpus, dft_command)
 
     call(qe_command, shell=True)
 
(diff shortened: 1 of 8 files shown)


dmclark17:
Python f string might be good? [Here](https://www.python.org/dev/peps/pep-0498/) is the PEP for it. It can be cleaner than the traditional string formatting

flare/dft_interface/qe_util.py
```diff
@@ -9,23 +9,32 @@
...
+            '{} -i {} > {}'.format(dft_loc, newfilename, dft_out)
+    else:
+        dft_command = \
+            '{} -npool {} -i {} > {}'.format(dft_loc, npool, newfilename,
+                                             dft_out)
 
-    return parse_dft_forces('pwscf.out')
+    if (no_cpus > 1):
+        # dft_command = 'mpirun -np {} {}'.format(no_cpus, dft_command)
+        dft_command = 'srun -n {} --mpi=pmi2 {}'.format(no_cpus, dft_command)
```
nw13slx: This line is the culprit for unit test failure. Should leave it to the user to decide.

Issue #26, opened 2019-09-11
To / From methods for Structure object & ASE Atoms


stevetorr:
@YuuuuXie  may have already done this, but
1. A static method for turning ASE Atoms objects intro FLARE Structures would help when generating structures using ASE,
2. A method for turning FLARE Structures into ASE Atoms objects would also be nice.

Both of these would enhance user pre-processing flexibility.

Issue #55, opened 2019-09-19
ASE interface unit test


YuuuuXie:
open a new branch, tasks:
1. calculator
2. otf - md
3. test interface of different dft calculators

YuuuuXie:
close via pull request #77

Issue #30, opened 2019-09-12
is_std_in_bound() method from OTF could be moved to a new file


stevetorr:
Related to issue 14: https://github.com/mir-group/flare/issues/14

Because the is_std_in_bound() function is a convenience method for OTF or (the in development) TrajectoryTrainer, it would eliminate redundancy to have it exist outside of the OTF class. If we end up splitting predict functions to a separate file, I think that this function would be a natural fit for there (given that it is to diagnose the result of a prediction on a structure).

YuuuuXie:
close via pull request #77

PR #77, opened 2019-09-27
Ase interface

YuuuuXie:
1. add *ase* folder, and remove `ase*.py` in *flare* folder
2. `ase` module includes flare calculator, otf engine and logger
3. different md engines are tested with otf (src: `otf_md.py`, unit test: `test_ase_setup/test_ase_otf.py` with QE)
4. add `.rst` for `mgp` and `ase` in *docs* folder
5. cleaned up `*.out` in *tests* folder since they are not used


flare/ase/calculator.py
```diff
@@ -0,0 +1,166 @@
...
+        return res
+
+def predict_on_structure(structure, gp_model):
+    res = []
+    for atom in range(structure.nat):
+        chemenv = AtomicEnvironment(structure, atom, gp_model.cutoffs)
+        res.append(predict_on_atom((structure, atom, gp_model)))
+    return res
+
+def predict_on_structure_en(strucutre, gp_model):
```
jonpvandermause: Input should be "structure"

flare/ase/calculator.py
```diff
@@ -0,0 +1,166 @@
...
+
+        train_size = len(self.gp_model.training_data)
+        rank_2 = np.min([1000, grid_params['grid_num_2'], train_size*3])
+        rank_3 = np.min([1000, grid_params['grid_num_3'][0]**3, train_size*3])
+        grid_params['svd_rank_2'] = rank_2
+        grid_params['svd_rank_3'] = rank_3
+       
+        self.mgp_model = MappedGaussianProcess(self.gp_model, grid_params, struc_params)
+
+def predict_on_atom(params):
```
jonpvandermause: These functions look pretty similar to the ones in predict.py, would it be possible to use those so that we don't repeat code?

flare/mgp/otf.py
```diff
@@ -18,7 +17,7 @@
     qe_input_to_structure, parse_qe_forces
 from flare import output
 
-class MFFOTF(OTF):
+class MGPOTF(OTF):
```
jonpvandermause: We now have three otf.py files in the repo -- mgp/otf, which inherits from flare/otf, and ase/otf, which is separate. If there are common methods to all three, we should try to import them from one place to minimize duplication.

YuuuuXie:
Jon: We now have three otf.py files in the repo -- mgp/otf, which inherits from flare/otf, and ase/otf, which is separate. If there are common methods to all three, we should try to import them from one place to minimize duplication.

YuuuuXie:
One common method that can be extracted out is `is_std_in_bound` which is sort of independent. MGPOTF is short because it inherits most of the methods from OTF. ase/otf is coupled with ase calculator, ase logger and ase md engines, so will be harder to be put out

jonpvandermause:
Yeah, is_std_in_bound would be good to have in one place (#30). We can either move it out of class in otf.py, or put it in predict.py.

YuuuuXie:
I can also have ase/otf to inherit from OTF somehow, since there're a lot of common attributes. Could we probably put `is_std_in_bound` in `util.py`? I think moving it in `predict.py` is a bit weird

jonpvandermause:
> Oh I used to use the same dir as you do (../test_files/pseudos) and never passed TRAVIS test. Then Lixin figured out TRAVIS did its unit test in the tests folder, so it runs python test_ase_setup/test_ase_otf.py. That's why I changed it to test_files/pseudos and then it passed unit test

Good catch. The test failed on my machine, but that's because I was running a different command than Travis. I switched it back.

anjohan:
Re: Travis testing directory.

I set Travis to run its tests from the tests directory. When I try to run pytest from the base directory, I get a lot of errors from trying to locate `./test_files/cp2k_input_1.in` etc.

YuuuuXie:
> Re: Travis testing directory.
> 
> I set Travis to run its tests from the tests directory. When I try to run pytest from the base directory, I get a lot of errors from trying to locate `./test_files/cp2k_input_1.in` etc.

I also tried running `pytest ./` in the *tests/* folder, but it works

YuuuuXie:
> Yeah, is_std_in_bound would be good to have in one place (#30). We can either move it out of class in otf.py, or put it in predict.py.

The latest commits removed the predict functions in ase/otf.py, and uses predict.py. Also removed `is_std_in_bound` in otf.py and ase/otf.py, and moved it to util.py 

This pull request can close #26 , #30 , #55

PR #82, opened 2019-10-01
Added draft of vasp_util.py for VASP trajectories and OTF

kylebystrom:
The new vasp_util.py file can edit VASP POSCAR files with new coordinates, read forces from a static calculation, and read trajectories from an ionic relaxation/MD calculation. This is a draft; I still need to write unit tests and documentation.


flare/dft_interface/vasp_util.py
```diff
@@ -0,0 +1,87 @@
+from pymatgen.io.vasp.inputs import Poscar
+from pymatgen.io.vasp.outputs import Vasprun
+from subprocess import call
+import numpy as np
+from flare import struc
```
stevetorr: I suggest keeping imports targeted to the individual object (as you did above for Pymatgen! :) ) so e.g. 
`from flare.struc import Structure`
Unless, did you do this to keep it clear the difference between a pymatgen structure and a flare structure?

flare/dft_interface/vasp_util.py
```diff
@@ -0,0 +1,87 @@
...
+	
+	currdir = os.getcwd()
+	if structure:
+		edit_dft_input_configurations("POSCAR", structure)
+	try:
+		os.chdir(calc_dir)
+	except Exception, e:
+		os.chdir(currdir)
+		call(vasp_cmd, shell=True)
+		raise e
```
stevetorr: Perhaps could change this to an if statement to see if calc dir and calc dir disagree, and then if so, use the chdir command to change to calc dir?

flare/dft_interface/vasp_util.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def parse_dft_forces_and_energy(vasprun: Union[str, Vasprun]):
+	if type(vasprun) == str:
+		vasprun = Vasprun(vasprun)
+	istep = vasprun.ionic_steps[-1]
+	return np.array(istep.forces), istep["electronic_steps"][-1]["e_0_energy"]
+
+
+def md_trajectory_from_vasprun(vasprun: Union[str, Vasprun]):
```
stevetorr: Maybe add an argument for ionic step skips?

stevetorr: Also I didn't know about the Union option for typehints. That's stylish, I like it!

flare/dft_interface/vasp_util.py
```diff
@@ -0,0 +1,87 @@
...
+	if type(vasprun) == str:
+		vasprun = Vasprun(vasprun)
```
stevetorr: If this condition is not met, and Vasprun is not of instance Vasprun, perhaps raise an exception?

stevetorr:
Will throw in some more comments later, but looks good so far!

flare/dft_interface/vasp_util.py
```diff
@@ -0,0 +1,87 @@
...
+		struc_lst.append(structure)
+
+	return struc_lst
+
+
+def md_trajectory_to_file(filename, structures: List[struc.Structure]):
+	f = open(filename, 'w')
+	dump([s.as_dict() for s in structures], f)
+	f.close()
+
```
stevetorr: This function is useful, but it's also VASP agnostic. Perhaps we could move it to another utility file?

flare/dft_interface/vasp_util.py
kylebystrom: Changed to the above suggestion

flare/dft_interface/vasp_util.py
kylebystrom: Yes, done

flare/dft_interface/vasp_util.py
kylebystrom: Added a reusable function to do this.

flare/dft_interface/vasp_util.py
kylebystrom: Added this by changing `for step in vasprun.ionic_steps` to `for step in vasprun.ionic_steps[::ionic_skip_steps]`

kylebystrom:
This should be ready to go now.

tests/test_flare_io.py
```diff
@@ -0,0 +1,16 @@
...
+
+def test_read_write_trajectory():
+	structures = md_trajectory_from_vasprun('test_files/test_vasprun.xml')
+	fname = 'tst_traj.json'
+	md_trajectory_to_file(fname, structures)
+	fstructures = md_trajectory_from_file(fname)
+	for s, f in zip(structures, fstructures):
+		assert np.isclose(s.forces, f.forces).all()
+		assert np.isclose(s.energy, f.energy).all()
+		assert np.isclose(s.positions, f.positions).all()
```
stevetorr: Good test!

tests/test_vasp_util.py
```diff
@@ -0,0 +1,92 @@
...
+                         ]
+                         )
+def test_structure_parsing(poscar):
+    structure = dft_input_to_structure(poscar)
+    pmg_struct = Poscar.from_file(poscar).structure
+    assert len(structure.species_labels) == len(pmg_struct)
+    assert (structure.cell == pmg_struct.lattice.matrix).all()
+    for i, spec in enumerate(structure.species_labels):
+        assert spec == pmg_struct[i].specie.symbol
+
```
stevetorr: Could you also add a line to assert the positions were parsed right?

flare/dft_interface/vasp_util.py
```diff
@@ -0,0 +1,128 @@
...
+        os.chdir(currdir)
+        raise e
+    
+
+def dft_input_to_structure(poscar: str):
+    """
+    Parse the DFT input in a directory.
+    :param vasp_input: directory of vasp input
+    """
+    return Structure.from_pmg_structure(Poscar.from_file(poscar).structure)
```
stevetorr: Can you ensure that this returns the cartesian coordinates every time since that's the FLARE convention?

stevetorr:
This looks great, Kyle! Can you address the two points above: 1. a small addition to the unit test, and 2. A guarantee that wherever in the code a FLARE structure is produced that the cartesian coordinates are returned?
Once that's set, feel free to pull :)

stevetorr:
Closes #87

flare/dft_interface/vasp_util.py
kylebystrom: This is addressed by the unit test for the from_pmg_structure function, but I also will address it in the `test_structure_parsing` function when I assert the correct coordinate parsing

tests/test_flare_io.py
kylebystrom: Thanks!

flare/dft_interface/vasp_util.py
kylebystrom: Actually, `test_struc.test_from_pmg_structure` doesn't guarantee the cartesian coordinates are read because it sets the unit cell to np.eye(3). So I copy-pasted the test and changed the lattice parameters so that it will fail if `from_pmg_structure` returns fractional coordinates.

tests/test_vasp_util.py
kylebystrom: Yup, added this in `test_structure_parsing`.

kylebystrom:
Okay, this is good to go now. Merging with master.

Issue #89, opened 2019-10-11
CPU overload when parallelization is on for both gp and prediction


nw13slx:
The training process can overload compute nodes when both gp and predict are parallelized.
This should not happen because the paralleled functions are never called in the same step.
But it could be because we use multithread and concurrence at the same time.

nw13slx:
one more observation. multiprocess.cpu_count() gives total number of cpu on the node, not the number requested by slurm. This also need to be fixed.

nw13slx:
The problem is in the gp.py predict function. It nests a parallelized function (set_L_alpha) in another parallelized function. I'll quickly fix this bug. But the multiprocessing model should be further improve.

PR #90, opened 2019-10-12
Bugfix/lixin/multiprocessing

nw13slx:
Temporarily fix the multithread nesting bug in #89 by removing the set_L_alpha sanity check from the parallelization.

More to do: 
1. replace the concurrent package in predict.py with multiprocessing. 
2. integrate the multiprocessing with pathos and slurm. So it can handle multinode environment.


YuuuuXie:
> Temporarily fix the multithread nesting bug in #89 by removing the set_L_alpha sanity check outside of the parallelization.
> 
> More to do:
> 
> 1. replace the concurrent package in predict.py with multiprocessing.
> 2. integrate the multiprocessing with pathos and slurm. So it can handle multinode environment.

(I have a branch for parallelization, maybe we can contribute to that branch for integration with pathos and slurm. )

YuuuuXie:
Is the module `md.py` in use? @jonpvandermause , we also need to move out the "predict" functions
I see there's also `concurrent` used there @nw13slx

nw13slx:
> Is the module `md.py` in use? @jonpvandermause , we also need to move out the "predict" functions
> I see there's also `concurrent` used there @nw13slx

Yes. Let's also remove those functions.

nw13slx:
close #89

PR #92, opened 2019-10-14
WIP: GP now takes opt_params argument

### LINKED ISSUE #91: Finer control over GP optimization parameters by instance
The TrajectoryTrainer and OTF classes contain calls to the `gp.train` method but do not pass in arguments that the `train` method can accept. In order to expose more control to the user, being able to set optimization parameters (such as `grad_tol` and `x_tol`) on an individual instance would be helpful, since it makes sense for those to be parameters associated with a given `GaussianProcess` instance. This could involve, for instance, re-factoring the train method to draw from an internal `opt_parameter_kwarg` dictionary which could be passed when instantiating the `GaussianProcess`.

stevetorr:
I decided to grandfather in `maxiter` and `opt_algorithm` as arguments because I thought that breaking everybody's scripts would be rude. I do think that they belongs in the  `opt_params` argument so we can add it in later if we want. Perhaps we could include a warning that `opt_algorithm` and `maxiter` will be deprecated with V 1.0 and to use `opt_params` instead if they are passed in as arguments.

Closes #91

### DIFF
--- a/flare/gp.py
+++ b/flare/gp.py
@@ -122,10 +122,16 @@ class GaussianProcess:
         return forces_np
 
     def train(self, output=None, opt_param_override: dict = None):
-        """Train Gaussian Process model on training data. Tunes the \
+        """
+        Train Gaussian Process model on training data. Tunes the \
 hyperparameters to maximize the likelihood, then computes L and alpha \
-(related to the covariance matrix of the training set)."""
+(related to the covariance matrix of the training set).
 
+        :param output: Output to write to
+        :param opt_param_override: Dictionary of parameters to override
+        instace's optimzation parameters.
+        :return:
+        """
         x_0 = self.hyps
 
         args = (self.training_data, self.training_labels_np,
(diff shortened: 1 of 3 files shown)


flare/gp.py
```diff
@@ -119,22 +121,39 @@ def force_list_to_np(forces: list):
...
+    def train(self, output=None, opt_param_override: dict = None):
+        """
+        Train Gaussian Process model on training data. Tunes the \
 hyperparameters to maximize the likelihood, then computes L and alpha \
-(related to the covariance matrix of the training set)."""
+(related to the covariance matrix of the training set).
 
+        :param output: Output to write to
+        :param opt_param_override: Dictionary of parameters to override
+        instace's optimzation parameters.
```
dmclark17: This could be a fun place to try out one of Python 3.8's new features! [TypedDicts](https://www.python.org/dev/peps/pep-0589/)! I actually don't know if this would be a good use case, but if we turned the `opt_param_override` into a TypedDict then Python's type checker could check the types in the dictionary

I am just looking for use cases of the new features of python so totally unnecessary to actually implement...

PR #93, opened 2019-10-14
Quick fix to suppress pymatgen warnings

kylebystrom:
The test_vasprun.xml is small and unconverged so it produces a warning. I suppressed this warning along with one that warns about a missing POTCAR file.


kylebystrom:
Merging this, just a small fix

PR #94, opened 2019-10-15
Feature/steven/gpfaimd checkpoints

stevetorr:
GPFAIMD now can leave checkpoints.
Added dedicated write files for the GP class to avoid code duplication.
Unit tested both of the above functionality.


stevetorr:
Oops, accidentally included the opt_params argument for GP in this commit... I'll undo that

stevetorr:
Note: I can rebase most of the above commits before pushing this to cut down on the small commits.

PR #99, opened 2019-10-21
Feature/steven/gpfa checkpoints

### LINKED ISSUE #98: Checkpoints from GPFA
Having checkpoints for GP from AIMD would be helpful when programs crash in the middle of training (for one reason or another). I have opened a pull request for this.

stevetorr:
Closes #98 .
Implements checkpoints in GPFA.

### DIFF
--- a/flare/gp_from_aimd.py
+++ b/flare/gp_from_aimd.py
@@ -4,17 +4,18 @@ trajectory. Contains methods to transfer the model to an OTF run /
 MD engine run.
 """
 import time
-from typing import List, Tuple
-from copy import deepcopy
-
+import json
 import pickle
 import numpy as np
 
+from typing import List, Tuple
+from copy import deepcopy
+
 from flare.output import Output
-from flare.struc import Structure, get_unique_species
+from flare.struc import Structure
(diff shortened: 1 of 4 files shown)


flare/gp_from_aimd.py
```diff
@@ -284,6 +299,11 @@ def train_gp(self):
                                self.gp.likelihood, self.gp.likelihood_gradient)
         self.train_count += 1
 
+        if self.checkpoint_interval \
```
nw13slx: cool. This is nice. Maybe we should also add a variable "last_step_written" to record when the gp was dumped. So it won't dump twice at the end of the run.

tests/test_gp.py
```diff
@@ -256,3 +262,29 @@ def test_serialization_method(two_body_gp, test_point):
     for d in [0, 1, 2]:
         assert np.all(two_body_gp.predict(x_t=test_point, d=d) ==
                       new_gp.predict(x_t=test_point, d=d))
+
+
+def test_load_and_reload(two_body_gp, test_point):
```
nw13slx: the unittest also looks good

PR #105, opened 2019-10-25
Yu/parallelize - close #83 & changed MGP

### LINKED ISSUE #83: update_L_alpha should be parallelized
set_L_alpha gives the option to compute the covariance matrix in parallel. It would be helpful to have the same option for update_L_alpha, especially for large GPs

YuuuuXie:
There are a few changes:
1. parallelized `update_L_alpha` function, which modified `gp.py` and `gp_algebra.py`
2. The previous MGP has the gp model as its attribute (i.e. `self.GP`), which is redundant, I eliminated this
3. The previous MGP builds up the spline functions when creating a `MappedGaussianProcess` object, I changed it to only creating an empty container (i.e. a spline class object without coefficients), and the coefficients can be fitted by calling `set_values` function. This change is for future's parallelization of MGP prediction


flare/ase/calculator.py
```diff
@@ -81,6 +82,51 @@ def get_forces_mgp(self, atoms):
         atoms.get_uncertainties = self.get_uncertainties
         return forces
 
+    def get_forces_mgp_par(self, atoms):
+        return self.get_forces_mgp_serial(atoms)
+#        comm = MPI.COMM_WORLD
```
nw13slx: why all the MPI command is commented out? maybe we can leave a note in the pull request that this is yet to be tested.

flare/gp_algebra.py
```diff
@@ -101,7 +101,7 @@ def get_ky_and_hyp_par(hyps: np.ndarray, training_data: list,
     # initialize matrices
     size = len(training_data) * 3
     k_mat = np.zeros([size, size])
-    hyp_mat = np.zeros([size, size, number_of_hyps])
+    hyp_mat = np.zeros([number_of_hyps, size, size])
```
nw13slx: nice. It should be noted in the pull request that the hyp_mat slicing has been changed

tests/test_ase_setup/otf_setup.py
```diff
@@ -12,7 +12,7 @@
 
 np.random.seed(12345)
 
-md_engines = ['VelocityVerlet', 'NVTBerendsen', 'NPTBerendsen', 'NPT']
+md_engines = ['VelocityVerlet'] #, 'NVTBerendsen', 'NPTBerendsen', 'NPT']
```
nw13slx: Is NVT available?

flare/ase/calculator.py
YuuuuXie: I think we might be able to use multiprocessing.shared_memory instead of mpi4py. I tested it, but there's some conflict inside of ASE's own parallel setting, calling QE and this MGP parallelization. 
The tricky part is we need to use `mpirun -n xx python **.py` if we use mpi4py, I think we must avoid this, and trying multiprocessing is safer. 

BTW, the `shared memory` feature of `multiprocessing` is new (only supported) in python 3.8....... We might need to transfer to this version....

tests/test_ase_setup/otf_setup.py
YuuuuXie: It is available. `otf_setup.py` is not the unit test file, the unit test file is `test_otf.py` which contains all MD engines. I modified it only because I used it in the documentation (tutorial)

PR #107, opened 2019-10-31
Ase interface - add stress & tested OTF parser

### LINKED ISSUE #86: ASE virial stress & test otf_parser for its output
1. calculate stress based on atomic forces
2. test otf_parser to parser the output file from ase-otf

YuuuuXie:
1. fixed some bugs
2. tested `otf_parser` on the ASE output file, unit test file is `test_ase_setup/test_parser_otf.py`
3. Add stress for MGP, using the same formulation as our lammps pair style. So the unit test is to match MGP stress prediction in python with the stress prediction from lammps mgp pair style. unit test file is `test_ase_setup/test_stress.py`
**Note:** since the lammps stress formula requires forces from pairs and triplets, if we want to implement stress for GP, we need to go into the kernel function, which is more complicated. So I just implemented for MGP now

close #86


docs/build/html/_static/searchtools.js
```diff
@@ -4,7 +4,7 @@
  *
```
jonpvandermause: We should add the docs/build directory to the .gitignore

PR #123, opened 2019-12-10
Added & unit tested store_dft_output

### LINKED ISSUE #122: Add option to store DFT runs in another folder during OTF calculation
For one reason or another, it may be desirable to hold onto the DFT data generated during an OTF run, from the inputs to the outputs (e.g. for VASP, charge density files and wave functions are stored in their own separate files). This is doubly true for expensive DFT calculations for large systems, for which the data may be desirable later.

One way to solve this would be to pass in an argument to OTF which takes a tuple of two elements: the first being a single file, or a list of files to copy to another directory, and the second argument being the directory. If this argument were used, after DFT calls, the files would be copied to the specified folder.

stevetorr:
Closes #122

### DIFF
--- a/flare/otf.py
+++ b/flare/otf.py
@@ -16,8 +16,8 @@ from flare.util import is_std_in_bound
 
 
 class OTF:
-    """Trains a Gaussian process force field on the fly while running a
-    molecular dynamics simulation.
+    """Trains a Gaussian process force field on the fly during
+        molecular dynamics.
 
     Args:
         dft_input (str): Input file.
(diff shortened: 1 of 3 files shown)


flare/otf.py
```diff
@@ -1,20 +1,23 @@
...
 from flare.dft_interface import dft_software
 from flare.output import Output
-import flare.predict as predict
 from flare.util import is_std_in_bound
 
 
 class OTF:
-    """Trains a Gaussian process force field on the fly.
+    """Trains a Gaussian process force field on the fly while running a
+    molecular dynamics simulation.
```
jonpvandermause: Slightly more succinct: "...on the fly during molecular dynamics."

flare/otf.py
```diff
@@ -61,6 +64,11 @@ class OTF:
             "srun".
         dft_kwargs ([type], optional): Additional DFT arguments. Defaults
             to None.
```
jonpvandermause: Unrelated, but would it be possible to add some more information about dft_kwargs to the docstring?

flare/otf.py
```diff
@@ -61,6 +64,11 @@ class OTF:
...
+        store_dft_output (Tuple, optional): After DFT calculations are called,
+            copy the file or files specified in the first element of the tuple
+            to a directory specified as the second element of the tuple.
+            Useful when DFT calculations are expensive and want to be kept
+            for later use.
     """
```
jonpvandermause: Is the first element of the tuple a list of strings? Or just a string? It looks like you've set it up to work with both, but it might be helpful to explicitly state that in the docstring. (You make this very clear in the constructor by giving the type as "Tuple[Union[str,List[str]],str]". Consider doing the same here, so that it shows up on RTD.)

flare/otf.py
```diff
@@ -212,6 +222,20 @@ def run(self):
                     if (self.dft_count-1) < self.freeze_hyps:
                         self.train_gp()
 
+                    # Store DFT outputs in another folder if desired
+                    # specified in self.store_dft_output
+                    if self.store_dft_output is not None:
+                        dest = self.store_dft_output[1]
+                        target_files = self.store_dft_output[0]
+                        now = datetime.now()
+                        dt_string = now.strftime("%Y.%m.%d:%H:%M:%S:")
```
jonpvandermause: This seems like a good way of labeling the output files. We should write the labeling convention in the docstring for the "run" method.

flare/otf.py
stevetorr: Done deal.

Issue #134, opened 2020-02-04
Error when trying to install via pip


smheidrich:
This happens both on my local machine as well as on our cluster if I try to install the package via pip:

```bash
$ pip install mir-flare
Collecting mir-flare
  Using cached https://files.pythonhosted.org/packages/7d/5e/9692de4a76b732072d552cc229672d41a33153dcd3f3a850dedda3475b47/mir-flare-0.0.5.tar.gz
    ERROR: Command errored out with exit status 1:
     command: /tmp/venv/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-c0v787oc/mir-flare/setup.py'"'"'; __file__='"'"'/tmp/pip-install-c0v787oc/mir-flare/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-c0v787oc/mir-flare/pip-egg-info
         cwd: /tmp/pip-install-c0v787oc/mir-flare/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-c0v787oc/mir-flare/setup.py", line 7, in <module>
        with open("requirements.txt", "r") as fh:
    FileNotFoundError: [Errno 2] No such file or directory: 'requirements.txt'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

What seems to happen is that ``requirements.txt`` wasn't included in the sdist published on PyPI. If I build the sdist myself, it doesn't get included there, either.

There is a very easy way to fix this, though, PR incoming...

smheidrich:
Fixed by 4685126c839c35a36811507cbaadfcedec59a3d3

PR #135, opened 2020-02-04
Add setuptools_scm as setup_requires dependency

smheidrich:
This causes anything under version control (particularly requirements.txt) to be included in sdists and wheels and should therefore make the package installable from PyPI again (cf. #134) the next time it gets published.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,7 @@ setuptools.setup(
     long_description_content_type="text/markdown",
     url="https://github.com/mir-group/flare",
     python_requires=">=3.6",
+    setup_requires="setuptools_scm",
     install_requires=dependencies,
     license="MIT",
     classifiers=[


anjohan:
Thank you for catching this and submitting a PR!

I was a bit surprised, since `pip install mir-flare==0.0.3` works and only the version has changed since, but the difference seems to be that I pushed wheels and not just sdists for the earlier versions.

From what I gather, setuptools-scm includes *all* files on git (docs, tests, etc.). I have instead added a MANIFEST.in which includes requirements.txt.

Please check if the newest version (0.0.6) fixes the issue (#134)! ("*It works on my machine.*")

smheidrich:
@anjohan Yep that works, I'll close this then, thanks! :+1:

PR #136, opened 2020-02-04
Ase interface - close #133

### LINKED ISSUE #133: Using the "trajectory" kwarg of OTF-MD engines leads to exceptions later on
According to [the documentation](https://flare.readthedocs.io/en/latest/flare/ase/otf_md.html#flare.ase.otf_md.otf_md), ``trajectory`` is a valid (even required, albeit "usually set to ``None``") key for the ``md_params`` dictionary that's a parameter of the ``otf_md()`` function. So I would expect to be able to let ASE's own MD machinery record a trajectory as usual by setting it to anything other than ``None``, e.g. a filename.

Here is a minimal reproducible example script in which I try to do just this (a bit long because FLARE requires a lot of boilerplate as far as I can tell):

<details><summary><b>flare_ase_trajectory_bug.py</b></summary>

```python
#!/usr/bin/env python3
import ase
from ase.calculators.emt import EMT
import numpy as np
from ase import units
from ase.md.velocitydistribution import (MaxwellBoltzmannDistribution,
                                         Stationary, ZeroRotation)
from flare.ase.otf_md import otf_md
from flare.ase.logger import OTFLogger

# Dummy system
a = ase.Atoms("H3")
a.positions[1:] += [[2,0,0], [2,2,0]]
a.set_cell(np.diag([10,10,10]))
a.set_pbc(True)

# Copied from https://flare.readthedocs.io/en/latest/tutorials/ase.html )
from flare.gp import GaussianProcess
from flare.ase.calculator import FLARE_Calculator
import flare.mc_simple as mc_simple

kernel = mc_simple.two_plus_three_body_mc
kernel_grad = mc_simple.two_plus_three_body_mc_grad
energy_force_kernel = mc_simple.two_plus_three_mc_force_en
energy_kernel = mc_simple.two_plus_three_mc_en

hyps = np.array([0.1, 1., 0.001, 1, 0.03])
two_cut = 4.0
three_cut = 4.0
cutoffs = np.array([two_cut, three_cut])
hyp_labels = ['sig2', 'ls2', 'sig3', 'ls3', 'noise']
opt_algorithm = 'LBFGS'

gp_model = GaussianProcess(kernel, kernel_grad, hyps, cutoffs,
                           energy_kernel=energy_kernel,
                           energy_force_kernel=energy_force_kernel,
                           hyp_labels=hyp_labels,
                           opt_algorithm=opt_algorithm, par=True)

flare_calc = FLARE_Calculator(gp_model, par=True, use_mapping=False)

# For minimal example, just use ASE's EMT calculator as dummy instead of DFT
dft_calc=EMT()

# MD and FLARE OTF stuff
a.set_calculator(flare_calc)

# set up OTF MD engine
md_params = {'timestep': 1 * units.fs,
             'trajectory': 'otf_md.traj', # XXX everything fine if this is None
             'dt': None}

otf_params = {'dft_calc': dft_calc,
              'init_atoms': [0],
              'std_tolerance_factor': 0.5,
              'max_atoms_added' : len(a.positions),
              'freeze_hyps': 0,
              'restart_from': None,
              'use_mapping': a.calc.use_mapping}

# intialize velocity
temperature = 600
MaxwellBoltzmannDistribution(a, temperature * units.kB)
Stationary(a) # zero linear momentum
ZeroRotation(a) # zero angular momentum

test_otf = otf_md('VelocityVerlet', a, md_params, otf_params)

# set up logger
test_otf.attach(OTFLogger(test_otf, a,
    logfile='otf_run.log', mode="w", data_in_logfile=True),
    interval=1)

# rescale temperature to 1200K at step 11
rescale_temp = [1200]
rescale_steps = [11]

# run otf
number_of_steps = 10
test_otf.otf_run(number_of_steps, rescale_temp, rescale_steps)
```
</details>

When I execute it, a regular ASE-compatible trajectory does in fact get written to disk, but after some more progress, this exception is raised:

```
Traceback (most recent call last):
  File "./flare_ase_trajectory_bug.py", line 80, in <module>
    test_otf.otf_run(number_of_steps, rescale_temp, rescale_steps)
  File "/home/smheidrich/flare/flare/ase/otf.py", line 135, in otf_run
    self.observers[0][0].add_atom_info(atom, self.stds[atom])
AttributeError: 'function' object has no attribute 'add_atom_info'
```

Putting ``None`` for ``trajectory`` instead makes it run fine.

I guess the issue is that ASE's own trajectory logger gets attached as the first observer to the OTF object, but FLARE relies on having its own logger as the first observer.

Initially I thought this would be a quick fix, but making this part of the code tolerate ASE's trajectory logger only leads to further problems in other places... so I thought I'd ask here before I waste much more time on this - perhaps it's an easy fix for you guys.

Everything was tested with the current revision on master (7afbd4bc8e5c2067189a2a714dd63f6fd29c0260).

YuuuuXie:
fixed the ASE logger issue reported in #133

### DIFF
--- a/flare/ase/calculator.py
+++ b/flare/ase/calculator.py
@@ -28,30 +28,30 @@ class FLARE_Calculator(Calculator):
         self.par = par
         self.results = {}
 
-    def get_property(self, atoms, property_name):
-        if property_name not in self.results.keys():
+    def get_property(self, name, atoms=None, allow_calculation=True):
+        if name not in self.results.keys():
             self.calculate(atoms)
-        return self.results[property_name]
+        return self.results[name]
 
 
     def get_potential_energy(self, atoms=None, force_consistent=False):
         if self.use_mapping:
             print('MGP energy mapping not implemented, temporarily set to 0')
-        return self.get_property(atoms, 'energy')
-
-
-    def get_forces(self, atoms):
-        return self.get_property(atoms, 'forces')
+        return self.get_property('energy', atoms)
+                                                 
+                                                 
+    def get_forces(self, atoms):                 
+        return self.get_property('forces', atoms)
 
 
     def get_stress(self, atoms):
         if not self.use_mapping:
             raise NotImplementedError("Stress is only supported in MGP")
-        return self.get_property(atoms, 'stress')
+        return self.get_property('stress', atoms)
 
 
     def get_uncertainties(self, atoms):
-        return self.get_property(atoms, 'stds')
+        return self.get_property('stds', atoms)
 
 
     def calculate(self, atoms):
--- a/flare/ase/calculator.py
+++ b/flare/ase/calculator.py
@@ -30,6 +30,8 @@ class FLARE_Calculator(Calculator):
 
     def get_property(self, name, atoms=None, allow_calculation=True):
         if name not in self.results.keys():
+            if not allow_calculation:
+                return None
             self.calculate(atoms)
         return self.results[name]
 
--- a/flare/ase/logger.py
+++ b/flare/ase/logger.py
@@ -50,7 +50,7 @@ class OTFLogger(MDLogger):
         self.logfile.write('\nhyperparameter optimization algorithm: ' +
                            gp_model.algo)
         self.logfile.write('\nuncertainty tolerance: {} times noise'.format(
-                           str(self.dyn.std_tolerance)))
+                           str(self.dyn.std_tolerance_factor)))
         self.logfile.write('\ntimestep (ps): {}'.format(self.dyn.dt/1000))
         self.logfile.write('\nnumber of frames: {}'.format(0))
         self.logfile.write('\nnumber of atoms: {}'.format(
--- a/flare/ase/otf.py
+++ b/flare/ase/otf.py
@@ -169,7 +169,7 @@ class OTF:
             curr_struc.stds = np.copy(self.stds)
             noise = self.atoms.calc.gp_model.hyps[-1]
             self.std_in_bound, self.target_atoms = is_std_in_bound(\
-                    noise, self.std_tolerance, curr_struc, self.max_atoms_added)
+                    noise, self.std_tolerance_factor, curr_struc, self.max_atoms_added)
 
             print('std in bound:', self.std_in_bound, self.target_atoms)
             #self.is_std_in_bound([])
--- a/flare/ase/otf.py
+++ b/flare/ase/otf.py
@@ -6,6 +6,7 @@ complete training module with OTF and MD.
 '''
 import os
 import sys
+import inspect
 from copy import deepcopy
 
 from flare.struc import Structure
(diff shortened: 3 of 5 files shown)


flare/ase/calculator.py
```diff
@@ -28,30 +28,30 @@ def __init__(self, gp_model, mgp_model=None, par=False, use_mapping=False):
         self.par = par
         self.results = {}
 
-    def get_property(self, atoms, property_name):
-        if property_name not in self.results.keys():
+    def get_property(self, name, atoms=None, allow_calculation=True):
+        if name not in self.results.keys():
```
smheidrich: This PR seems to fix my issue (thanks!) but makes the code run insanely slowly compared to having ``trajectory=None`` in my tests. I think ``allow_calculation`` should be actually handled here [the way ASE does it](https://gitlab.com/ase/ase/-/blob/f9c6203b03620d4231c1c147c170c9196a1f14d1/ase/io/trajectory.py#L427-429):

```suggestion
        if name not in self.results.keys():
            if not allow_calculation:
                return None
```

⇒ Fast again

YuuuuXie: Thanks for finding this issue and giving the suggestion. A new commit has been made on this `allow_calculation` change.

smheidrich:
Thanks again for fixing this, I think there is one more issue with this PR: The resulting trajectory file has about twice as many frames as there were MD steps - I guess this is because, as far as I understand, ``call_observers`` is called a second time whenever there is a DFT calculation, whereas in ASE, it's called only once per MD step.

YuuuuXie:
Thanks for pointing out this. Actually we were not using ASE trajectory for doing things, so we did not consider the compatibility with it. Since we have our own `observer`/`logger`, we have to record the positions and forces from DFT calculations, so as to collect training data, that's why `call_observers` is also used in DFT call. 

I will open a new issue for this enhancement, i.e. to make our log file better compatible with ASE primitive log files, or even merge some of them. 

For your ASE trajectory file, a temporary solution is to remove the DFT frames in post processing. You can look at the folder `otf_data` generated by your otf run. In the file `dft_positions.xyz` all the frames for DFT calculation are collected, you will see lines like `Frame: x` meaning at step x, GP gives a force prediction, and then DFT is also called to make a force prediction for the same frame. 

For example, in the toy example you gave (in #133 ), you can run the following to remove the DFT frames in your trajectory file (the `new_trj` is the new trajectory file with DFT frames removed):
```
from ase.io.trajectory import Trajectory

# extract dft frame number from file
dft_file = open('otf_data/dft_positions.xyz')
dft_frames = []
for line in dft_file.readlines():
    if 'Frame' in line:
        dft_frames.append(int(line.split()[-1]))

# ase trajectory
trj = Trajectory('otf_md.traj')
new_trj = Trajectory('otf_md_new.traj', mode='w')
j = 0
for i in range(len(trj)):
    md_step = len(new_trj)
    if md_step == dft_frames[j]: # the current step is dft 
        j += 1
        continue
    else:
        new_trj.write(trj[i])
```

smheidrich:
@YuuuuXie Thanks a lot for going to the trouble of writing a workaround! I'll use this for the time being.

Making FLARE's OTF-MD stuff more compatible with ASE's MD classes in general would be great, too, of course.

PR #146, opened 2020-02-24
Merge/mc simple+development

nw13slx:
This branch merges @AldoGl and @ClaudioZeni 's development of the many body kernel.

Aside from their original kernel functions, following changes are implemented to accommodate the new kernel.
1. cutoff_2 and cutoff_3 variables in Environment class is replaced by a unify cutoffs vector
2. test_env and test_gp_from_aimd are changed for this replacement.
3. kernels.utils.str_to_kernel_set are updated to allow using many body kernel with gp. (example string: "2+3+mb+mc")
4. dummy mc_simple in the main tree is removed
5. test_mc_kernels is appended with new tests and renamed to mb_mc_kernels
6. some None-value members are added in Environment for no manybody kernel environment
7. the many body kernel is also blended with mc_sephyps, to allow different hyper-parameters for two-body and three-body. But it still uses the same sigm and lsm for all many body interactions.


flare/kernels/utils.py
```diff
@@ -98,19 +106,37 @@ def from_mask_to_args(hyps, hyps_mask: dict, cutoffs):
...
+        if (n2b == 0) and (n3b == 0):
+            raise NameError("Hyperparameter mask missing nbond and/or"
+                            "ntriplet key")
+
+        return (np.array(cutoffs), hyps_mask['nspec'], np.array(hyps_mask['spec_mask']),
+                n2b, np.array(bond_mask), n3b, np.array(triplet_mask),
+                np.array(sig2), np.array(ls2), np.array(sig3), np.array(ls3))
+    elif (len(cutoffs)==3):
+        sigm = None
+        lsm = None
```
AldoGl: these should probably be moved outside of the if statement

nw13slx: ok. just changed it.

PR #149, opened 2020-02-25
give the option to print hyperparameters to console

### LINKED ISSUE #147: Option to print hyperparameters directly
In the current implementation, when the hyperparameters of a GP are optimized, an output object needs to be created in order for the progress of the optimization to be recorded; this creates a few extraneous otf files, e.g. otf_run-f.xyz and otf_run-std.xyz.

As a near term fix, the user can be given the option to print the current hyperparameters directly, which is convenient when running simple tests. We'll eventually want to decouple hyperparameter optimization from otf runs entirely, since they're logically distinct.

jonpvandermause:
Closes #147 

Gives the option to print hyperparameters, likelihood, and likelihood gradient during hyperparameter optimization. Defaults to false.

### DIFF
--- a/flare/gp_algebra.py
+++ b/flare/gp_algebra.py
@@ -174,6 +174,7 @@ def get_ky_and_hyp(hyps: np.ndarray, training_data: list,
 
     return hyp_mat, ky_mat
 
+
 def get_ky_mat_par(hyps: np.ndarray, training_data: list,
                    kernel, cutoffs=None, ncpus:int =None):
     """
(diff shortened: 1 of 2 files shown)


flare/gp_algebra.py
```diff
@@ -504,4 +505,9 @@ def get_neg_like_grad(hyps, training_data: list,
     if output is not None:
         output.write_hyps(None, hyps, None, like, like_grad, name="hyps")
 
+    if print_progress:
```
nw13slx: maybe merge this condition block to the one above and make it "elif print_progress" ? just to reduce redundancy

jonpvandermause: Hmm I think we should leave open the option to do both, i.e. to both print to console to check the progress in real time, and to also record the progress in an output file so that it's stored.

Issue #127, opened 2019-12-11
issues with output.py


jonpvandermause:
- positions, forces, and velocities have a variable number of decimal places, e.g.:
![image](https://user-images.githubusercontent.com/37220876/70649124-7ff96680-1c1a-11ea-941b-ead754946231.png)

- simulation time is only reported up to two decimal places, giving plots that look like this:
![image](https://user-images.githubusercontent.com/37220876/70648829-edf15e00-1c19-11ea-90e6-4cd5fd4ba479.png)

- after a DFT calculation, the wall time is reported in scientific notation:
![image](https://user-images.githubusercontent.com/37220876/70648925-1da06600-1c1a-11ea-8c2e-52d067bec8d6.png)
Everywhere else it's reported up to two decimal places:
![image](https://user-images.githubusercontent.com/37220876/70649010-4c1e4100-1c1a-11ea-9753-2de558c38334.png)

- there are issues with tabbing:
![image](https://user-images.githubusercontent.com/37220876/70649056-60623e00-1c1a-11ea-81bd-33b967038ea2.png)

- otf_run-f.xyz and otf_run-stat.dat are empty during OTF training runs

- the same frame can be repeated multiple times in otf_run.xyz files:
![image](https://user-images.githubusercontent.com/37220876/70651626-2b0c1f00-1c1f-11ea-8831-e0de8cfef4e7.png)

- there is redundancy in the record_state method of otf:
![image](https://user-images.githubusercontent.com/37220876/70653071-c7cfbc00-1c21-11ea-8800-2136bc0e1fe6.png)
In particular, write_xyz_config is called within write_md_config, leading to duplication in the output files.

- there are unnecessary print statements in write_md_config:
![image](https://user-images.githubusercontent.com/37220876/70653298-34e35180-1c22-11ea-9f77-c51cdc7385a5.png)

- it's unclear why only the std output file is updated in write_md_config:
![image](https://user-images.githubusercontent.com/37220876/70653325-4593c780-1c22-11ea-8cf7-e7a2a79c71b9.png)

Issue #148, opened 2020-02-25
OTF and GPFA have duplicated code


jonpvandermause:
We can create a parent class, or perhaps better, just write independent functions that both classes call.

nw13slx:
I'll vote for a parent class. So users can write their own training protocol by setting up a children class. And it's easier for communication as well.

PR #152, opened 2020-02-26
Add custom force module and otf helper functions; remove xyz logs

### LINKED ISSUE #151: Users should have the option to give OTF a custom force module
This increases the flexibility of the program, and makes it easier to run simple OTF tests using fast force calculators.

### LINKED ISSUE #153: Helper functions needed for creating supercells and assigning initial velocities
These could be placed in the md.py module or in a new file (e.g. otf_helpers.py).

jonpvandermause:
Closes #151 by giving users the option to give the OTF trainer a custom force module.

Closes #153 by adding velocity initialization helper functions to md.py and supercell construction helper functions to util.py.

Because of the various issues with the .xyz logs described in #127, these output files have been removed for now.


flare/md.py
```diff
@@ -34,3 +35,61 @@ def calculate_temperature(new_pos, structure, dt, noa):
...
+    Args:
+        noa (int): Number of atoms in the system.
+        temperature (float): Temperature of the system.
+        mass (float): Mass of each particle in amu.
+
+    Returns:
+        np.ndarray: Particle velocities, corrected to give zero center of mass motion.
+    """
+
+    mass_md = mass * 0.000103642695727
```
dmclark17: What is this number? Can you please make it a named variable?

flare/md.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+from typing import List
```
dmclark17: Can I get a PEP8 check on these import statements?

flare/otf.py
```diff
@@ -83,10 +89,10 @@ def __init__(self, dft_input: str, dt: float, number_of_steps: int,
...
                  max_atoms_added: int = 1, freeze_hyps: int = 10,
                  rescale_steps: List[int] = [], rescale_temps: List[int] = [],
-                 dft_softwarename: str = "qe",
-                 no_cpus: int = 1, npool: int = None, mpi: str = "srun",
-                 dft_kwargs=None,
-                 store_dft_output: Tuple[Union[str,List[str]],str] = None):
+                 dft_softwarename: str = "qe", no_cpus: int = 1,
+                 npool: int = None, mpi: str = "srun", dft_kwargs=None,
+                 store_dft_output: Tuple[Union[str, List[str]], str] = None,
+                 custom_module=None):
```
stevetorr: How about we re-name "dft_softwarename" to "force_source", and let it take an object or a string as an argument? If the string is 'qe' or 'vasp' or 'cp2k' then it uses the `dft_software` dictionary to link to the correct module, and if it's of type `object`, then it becomes to the `custom_module` case. Would help to cut down on bloat. We could eventually also, with some work, allow for `trajectory` as an option, which could be a first step towards closing #148 .

flare/md.py
stevetorr: Added comments explaining them.

flare/md.py
```diff
@@ -34,3 +35,64 @@ def calculate_temperature(new_pos, structure, dt, noa):
...
+        mom_curr = rand_vel * mass_md
+        mom_tot += mom_curr
+
+    # Correct momentum, remove center of mass motion
+    mom_corr = mom_tot / noa
+    for n, mass in enumerate(masses):
+        mass_md = mass * 0.000103642695727
+        velocities[n] -= mom_corr / mass_md
+
+    return velocities
```
stevetorr: This all looks good. It might be nice to have a `set_temperature(T, dt)` method on `Structure` in `struc.py` which sets the `prev_position`s accordingly. In which case, these could be moved to `util` so as to not induce a dependency between `struc.py` and `md.py`.

flare/otf.py
```diff
@@ -74,6 +74,12 @@ class OTF:
             single file name, or a list of several. Copied files will be
             prepended with the date and time with the format
             'Year.Month.Day:Hour:Minute:Second:'.
+        custom_module: A custom module that can be used in place of the DFT
```
stevetorr: Just make sure to update this guy and to also scan for unit tests which might try to pass in 'dft_softwarename' -- as a side note, does your IDE have a `change name` function or similar which checks for usage within the same project? It can be quite nice to use!

stevetorr: Update: didn't see that you had already done so below.  Sweet!

flare/md.py
jonpvandermause: I moved the velocity helpers to util.py to set this up, but we can leave the actual modification of struc for another issue/PR.

PR #161, opened 2020-04-06
Removes wildcard exception for L-BFGS-B training 

### LINKED ISSUE #160: Wildcard Exception Handling in During Training
The L-BFGS-B minimization algorithm is in a wildcard try-catch block. When I try to kill the training with a keyboard interrupt, it just goes to the next training algorithm. I think a keyboard interrupt should end the program; to fix this we could either find out what exceptions minimize will throw and specifically catch them (smart), or specially catch the keyboard interrupt exception and exit the program (hacky).

Thoughts?

dmclark17:
Closes #160

[This](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.LinAlgError.html#numpy.linalg.LinAlgError) seems like the exception the comment was referring to. 

Right now these lines are not tested, but if we want to boost that code coverage I think we need to include pytest mock to write the unit tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ cache:
 before_install:
     - sudo apt update
     - sudo apt install -y quantum-espresso tree
-    - pip install codecov pytest pytest-cov
+    - pip install codecov pytest pytest-cov pytest_mock
     - pwd
     - tree
     - wget http://folk.uio.no/anjohan/lmp
--- a/flare/gp.py
+++ b/flare/gp.py
@@ -263,7 +263,7 @@ class GaussianProcess:
                                options={'disp': False, 'gtol': grad_tol,
                                         'maxls': line_steps,
                                         'maxiter': self.maxiter})
-            except:
+            except np.linalg.LinAlgError:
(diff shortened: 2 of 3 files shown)


dmclark17:
I added a unit test for this functionality because I was waiting for something to compile. It required some restructuring of the gp test file because I was modifying the fixture. On that note, one of the other tests also modified the GP fixture and then another test relied on that modification. I do not think this is good testing form. In my opinion all the tests should be able to be run independently of each other - even if they are integration tests and not true unit tests. I did not fix this, but it should probably be addressed at some point.

stevetorr:
I was writing a GP test a while ago and ran into this issue based on inter-test dependency. The principle as you articulated makes sense (I'll keep it in mind for my own future test-writing).

dmclark17:
:+1:

I believe this one is because one tests adds data to the GP and another tests asserts the length of the feature vector- it makes sense to have this as a test as the feature vector length should increase when data is added. We could just combine these tests I believe, but not a big deal in this case

dmclark17:
I think these tests add too much complexity for the additional 1 line of code coverage they add. I'm going to remove them and redo the pull request with just the actual code change

PR #162, opened 2020-04-08
[WIP] Development branch with many body feature and energy mapping

nw13slx:
- Energy mapping, and enhanced the ase interface @YuuuuXie 
- Manybody kernel by @AldoGl  @ClaudioZeni 
- much shorter unit tests
- easier gp initialization
- more structured interface for different kernels
- compatibility with ZAPPA @stevetorr 
- Faster GP by using copy-on-right with global variable for multiprocessing. (For 188 atomic environments and mc_sephyps kernels, it takes about 0.83 second to predict forces for atomic environment, and 5.7 seconds for set_L_alpha() with 32 Skylake cores. )


YuuuuXie:
I'm thinking we can probably have cross review for example, I can review the many-body part since I've never seen/used the code before. 

And @nw13slx @jonpvandermause @stevetorr, you can pick up a part of change that is not written by yourself to review

jonpvandermause:
> I'm thinking we can probably have cross review for example, I can review the many-body part since I've never seen/used the code before.
> 
> And @nw13slx @jonpvandermause @stevetorr, you can pick up a part of change that is not written by yourself to review

Good idea! I'll review the energy mapping and ase interface.

flare/env.py
```diff
@@ -190,6 +200,97 @@ class to allow for njit acceleration with Numba.
     return bond_array_2, bond_positions_2, etypes
 
 
+@njit
+def get_2_body_arrays_ind(positions, atom: int, cell, cutoff_2: float, species):
```
YuuuuXie, line 204: `get_2_body_arrays` and `get_2_body_arrays_ind` are redundant. The latter is used in the `get_m_body_arrays`, and should be changed into the similar style as `get_3_body_arrays`, i.e. search arrays from a shorter list (if we presume that 2-body cutoff is larger than many body)

tests/test_kernels.py
```diff
@@ -644,3 +664,293 @@ def test_grad_mask_function():
...
     assert (from_grad_to_mask(grad, hyps_mask) == [1, 4]).all()
+
+
+# -----------------------------------------------------------------------------
+#                              test many body kernels
+# -----------------------------------------------------------------------------
+
+
+def test_many_body_force():
+    """Check that the analytical force kernel matches finite difference of
```
YuuuuXie, line 675: This chunk of code can be simplified via loops....

tests/test_mc_mb_kernel.py
```diff
@@ -0,0 +1,461 @@
...
+                 original)/delta
+        print("numerical gradients", hgrad)
+        assert(np.isclose(grad_test[1][i], hgrad, atol=tol))
+
+# -----------------------------------------------------------------------------
+#                              test many body kernels
+# -----------------------------------------------------------------------------
+
+
+def test_many_body_force():
```
YuuuuXie, line 168: same as test_kernels.py, can be simplified

tests/test_mgp_unit_ff.py
```diff
@@ -0,0 +1,221 @@
+import numpy as np
```
YuuuuXie, line 1: mgp_en.py and mgp.py are quite duplicated. I'll consider merging them, and set up options for "energy map" or "force map"

flare/env.py
AldoGl: Yes, we did not have a deep knowledge of the code so we chose to duplicate the function to risk as little as possible and for lack of time,  sorry for this!

YuuuuXie: No worry! The many-body code is very clean and well-organized overall.

tests/test_kernels.py
nw13slx: this will be address in pull #166

PR #170, opened 2020-05-14
Enable multiple cutoff for two-, three- and manybody kernel

nw13slx:
Kernels and environments:
- new kernels that enable multiple group of hyperparameters
- AtomEnvironment is modified to allow using different cutoffs to define the bond, triplet and coordination number arrays
- conditions insert in all kernels to skip the zero terms faster
- helper functions in kernel/utils are updated accordingly

Unit tests
- cleaner unit tests for mc_sephyps
- more thorough tests for environment
- clean up fake_gp

User interface:
- HyperParameterMasking class to help easy definition. Move the hyper-parameter mask relevant checks and conversion from gp to mask_helper module.


flare/env.py
```diff
@@ -32,40 +76,84 @@ def __init__(self, structure: Structure, atom: int, cutoffs, sweep=1):
...
+        #         "mb cutoff has to be larger than mb cutoff"
 
         self.compute_env()
 
+    def setup_mask(self):
+
+        self.scalar_cutoff_2, self.scalar_cutoff_3, self.scalar_cutoff_mb, \
+            self.cutoff_2b, self.cutoff_3b, self.cutoff_mb, \
+            self.nspecie, self.n2b, self.n3b, self.nmb, self.specie_mask, \
+            self.bond_mask, self.cut3b_mask, self.mb_mask = \
```
stevetorr, line 97: It's good practice to make sure all the attributes that the class could have are defined in the __init__ function, even if they are populated later. I see what you have in mind by wrapping it up in a tiny function but I think they can stay in the __init__.

flare/env.py
```diff
@@ -32,40 +76,84 @@ def __init__(self, structure: Structure, atom: int, cutoffs, sweep=1):
...
+        # assert self.scalar_cutoff_mb <= self.scalar_cutoff_2, \
+        #         "mb cutoff has to be larger than mb cutoff"
 
         self.compute_env()
 
+    def setup_mask(self):
+
+        self.scalar_cutoff_2, self.scalar_cutoff_3, self.scalar_cutoff_mb, \
+            self.cutoff_2b, self.cutoff_3b, self.cutoff_mb, \
+            self.nspecie, self.n2b, self.n3b, self.nmb, self.specie_mask, \
```
stevetorr, line 96: On a separate note, can we more verbosely name n2b and n3b?

stevetorr: n2b_groups? n3b_groups?
n2body_groups? n3body_groups?

nw13slx: omg, that's a long name. Sure. we can. will do it in the refactor branch...

tests/test_otf.py
```diff
@@ -33,7 +33,7 @@ def cleanup(software="qe", casename="h2_otf_cp2k"):
     for f in glob.glob(f"*{software}.out"):
         os.remove(f)
 
-    if (software is 'qe'):
+    if (software == 'qe'):
```
stevetorr, line 36: The parentheses are redundant :) (And Pep8 doesn't like that)

flare/env.py
nw13slx: yes. I sort of implicitly do it. But it's more clear to set it up front in __init__

tests/test_otf.py
nw13slx: ok. my autopep8 didn't manage to get rid of it...

flare/kernels/kernels.py
```diff
@@ -277,6 +277,33 @@ def q_value_mc(distances, r_cut, ref_species, species, cutoff_func, q_func=coord
...
+    Return:
+        float: the value of the many-body descriptor
+    """
+
+    q = 0
+
+    for i in range(len(distances)):
+        if species[i] == ref_species:
+            q_, _ = q_func(distances[i], 0, r_cut, cutoff_func)
+            q += q_
```
stevetorr, line 302: Never saw the convention of temporary variables being underlined e.g. `q_`, but I like it! :)

flare/env.py
stevetorr: `n2b_groups` is my vote ;)

flare/env.py
stevetorr: See Zen of Python, PEP20, Line 2:
https://www.python.org/dev/peps/pep-0020/
`Explicit is better than implicit.` :)

tests/test_mask_helper.py
```diff
@@ -0,0 +1,137 @@
...
+    print(hm['hyps'])
+    print("obtain test hm", hm)
+
+    pm1 = HyperParameterMasking.from_dict(hm, verbose=True)
+    print("from_dict")
+    hm1 = pm1.generate_dict()
+    print(hm['hyps'])
+    print(hm1['hyps'][:33], hm1['hyps'][33:])
+
+    dumpcompare(hm, hm1)
```
stevetorr, line 137: It occurs to me that it might also make sense to try up a few wrong or incorrect setups, and with an exception catch, also test that check_instantiation correctly vetoes them. To unit test the tests in a way...

tests/test_mc_sephyps.py
```diff
@@ -4,219 +4,436 @@
...
-    cutoffs = np.ones(3, dtype=np.float64)
-    delta = 1e-8
-    env1_1, env1_2, env1_3, env2_1, env2_2, env2_3 = \
-        generate_envs(cutoffs, delta)
-
-    # set hyperparameters
     d1 = 1
     d2 = 2
     tol = 1e-4
+    cell = 1e7 * np.eye(3)
```
stevetorr, line 27: Wow! That's a big cell!

flare/ase/calculator.py
```diff
@@ -158,9 +163,11 @@ def build_mgp(self, skip=True):
...
-                        container_only, self.gp_model, lmp_file_name)
-
+        self.mgp_model = MappedGaussianProcess(grid_params,
+                                               struc_params,
+                                               map_force=map_force,
+                                               GP=self.gp_model,
+                                               mean_only=mean_only,
+                                               container_only=container_only,
+                                               lmp_file_name=lmp_file_name,
+                                               n_cpus=n_cpus)
```
stevetorr, line 173: I'm in general a huge fan of passing in as much as possible via keyword arguments this way for non user-facing functions; helps to make sure that the function is robust against later re-factoring which changes the names of the arguments.

flare/kernels/map_3b_kernel.py
```diff
@@ -0,0 +1,654 @@
...
+    return three_body_mc_en_force_jit(env1.bond_array_3, env1.ctype,
+                                      env1.etypes,
+                                      env1.cross_bond_inds,
+                                      env1.cross_bond_dists,
+                                      env1.triplet_counts,
+                                      c2, etypes2,
+                                      r1, r2, r12,
+                                      d1, sig, ls, r_cut, cutoff_func) / 3
+
+def three_body_mc_en_force_sephyps(env1, r1, r2, r12, c2, etypes2,
```
stevetorr, line 148: Phew... Multicomponent, energy/force, and separate hyps. Definitely a sign modularity is needed :)

flare/kernels/map_3b_kernel.py
```diff
@@ -0,0 +1,654 @@
...
+#                     if (c1 == ej2):
+#                         if (ei1 == ej1) and (ei2 == c2):
+#                             kern += three_body_en_helper(ci1, ci2, r13, r22,
+#                                                          r31, fi, fj, fdi, ls1,
+#                                                          ls2, sig2)
+#                         if (ei1 == c2) and (ei2 == ej1):
+#                             kern += three_body_en_helper(ci1, ci2, r12, r23,
+#                                                          r31, fi, fj, fdi, ls1,
+#                                                          ls2, sig2)
+#     return kern
```
stevetorr, line 354: Can this comment block be removed?

flare/kernels/mc_mb_sepcut.py
```diff
@@ -0,0 +1,391 @@
...
+
+        # compute kernel only if central atoms are of the same species
+        if c1 == c2:
+            k12 = k_sq_exp_double_dev(q1, q2, sig[mbtype1], ls[mbtype1])
+        else:
+            k12 = 0
+
+        # initialise arrays of many body descriptors and gradients for the neighbour atoms in
+        # the two configurations
+        # Loop over neighbours i of 1st configuration
```
stevetorr, line 81: Leaving the comment here but, in general, really nice comments! :)

flare/kernels/mc_sephyps.py
```diff
@@ -65,161 +103,318 @@
...
     two_term = \
         two_body_mc_force_en_jit(env1.bond_array_2, env1.ctype, env1.etypes,
                                  env2.bond_array_2, env2.ctype, env2.etypes,
-                                 d1, sig2, ls2, r_cut_2, cutoff_func,
+                                 d1, sig2, ls2, cutoff_2b, cutoff_func,
                                  nspec, spec_mask,
                                  bond_mask) / 2
 
+    if (ncut3b == 0):
+        tbmcj = three_body_mc_force_en_jit
```
stevetorr, line 325: I think we can afford to make this more verbose :) `tb_mc_jit_func?

flare/mgp/mgp.py
```diff
@@ -5,117 +5,184 @@
...
         self.n_sample = n_sample
+        self.grid_params = grid_params
+        self.struc_params = struc_params
+        self.hyps_mask = None
+        self.cutoffs = None
+
+        # arg_dict = inspect.getargvalues(inspect.currentframe())[3]
+        # del arg_dict['self'], arg_dict['GP']
+        # self.__dict__.update(arg_dict)
+        self.__dict__.update(grid_params)
```
stevetorr, line 109: I still have mixed feelings about this convention for loading in grid param values. While it does make the grid params a little uglier to define them explicitly, I do think it's helpful when they are referenced elsewhere, and it also allows IDEs to infer other things about the attributes. I think there's a broader conversation to be had with @YuuuuXie about how perhaps the role of grid_params could be changed-- maybe as a part of the refactory.

flare/mgp/mgp.py
```diff
@@ -5,117 +5,184 @@
...
+            self.cutoffs = GP.cutoffs
+            self.hyps_mask = GP.hyps_mask
+
+            self.bodies = []
+            if "two" in GP.kernel_name:
+                self.bodies.append(2)
+                self.kernel2b_info = get_2bkernel(GP)
+            if "three" in GP.kernel_name:
+                self.bodies.append(3)
+                self.kernel3b_info = get_3bkernel(GP)
```
stevetorr, line 129: Lovely!

flare/mgp/mgp.py
```diff
@@ -5,117 +5,184 @@
...
         self.build_bond_struc(struc_params)
         self.maps_2 = []
         self.maps_3 = []
-        self.build_map_container()
+        self.build_map_container(GP)
+        self.mean_only = mean_only
 
-        if ((not container_only) and (GP is not None) and (len(GP.training_labels_np) > 0)):
+        if not container_only and (GP is not None) and \
+                (len(GP.training_data) > 0):
```
stevetorr, line 138: I see the` autorun` `bool` variable is gone. I won't die on this hill, but, one advantage of allowing the user to call the run manually means that some init variables could be come arguments passed into the `build_map` function. Right now `MappedGaussianProcess` is pretty lean, but something to keep in mind in the future.

flare/mgp/mgp.py
```diff
@@ -5,117 +5,184 @@
...
-    def build_map_container(self):
+    def build_map_container(self, GP=None):
         '''
-        construct an empty spline container without coefficients
+        construct an empty spline container without coefficients.
         '''
+
+        if (GP is not None):
+            self.cutoffs = GP.cutoffs
+            self.hyps_mask = GP.hyps_mask
```
stevetorr, line 148: Love this economy!

flare/kernels/mc_sephyps.py
nw13slx: hahaha . Sure. because the variable is used immediately in the next line. So I just make it very short...

flare/env.py
nw13slx: This is implemented in the new factory branch

flare/kernels/map_3b_kernel.py
nw13slx: It's left there for the next step (force mapping). It'll be revived in the factory branch

flare/mgp/mgp.py
nw13slx: I agree. As we discuss last time, we will replace all this init with explicit function. That make me wonder, whether we should do the same to hyps_mask, at least for the GP initialization. But that will make everything super lengthy...

YuuuuXie: @stevetorr Absolutely I'm gonna re-organize this part. Actually some parameters in grid_params will be removed. 

@nw13slx If you are gonna change GP, I can wait until you get GP done, and see if we need to follow the same style with MGP

flare/mgp/mgp.py
YuuuuXie: I delete that just because it is duplicated with `container_only`, so every time the user needs to specify `container_only` and `autorun` simultaneously. As for the variable name, I just feel that `container_only` is more explicit so I kept that one

flare/ase/calculator.py
YuuuuXie: As we discussed, this part need re-considering. I might just change it to a "rebuild" method

flare/mgp/mgp.py
stevetorr: Ahhh yes, you're right, makes sense!

PR #172, opened 2020-05-18
Feature/steven/gpfa parser

stevetorr:
Adds:
- `parse_trajectory_trainer_output` to gp_from_aimd.py which allows for easy parsing on both a frame-by-frame level as well as on the GP level.
- A unit test for the parser, and a test file for it to run on.

Changes:
- Certain output.py functions now have their arguments changed to allow optional input. This should change nothing about the way they are used in other scripts, but makes for cleaner GPFA output parsing.
- Minor cosmetic change to GaussianProcess' training_statistics method.

### DIFF
--- a/flare/gp.py
+++ b/flare/gp.py
@@ -988,7 +988,7 @@ class GaussianProcess:
                                                     env.atom]))
 
         # Summarize the relevant information
-        data['species'] = set(present_species)
+        data['species'] = list(set(present_species))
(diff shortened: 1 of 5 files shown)


stevetorr:
Flip-flopped the master / dev base because I was a little confused as to why suddenly so many commits showed up. I missed that Jon recently pulled energy GP work onto development. I'll cherry pick this onto master instead.

PR #173, opened 2020-05-18
Added GPFA parser function to gp_from_aimd

stevetorr:
Adds:

- `parse_trajectory_trainer_output` to `gp_from_aimd.py` which allows for easy parsing on both a frame-by-frame level as well as on the GP level.
- A unit test for the parser, and a test file for it to run on.
Changes:

- Certain `output.py` functions now have their arguments changed to allow optional input. This should change nothing about the way they are used in other scripts, but makes for cleaner GPFA output parsing.
- Minor cosmetic change to GaussianProcess' training_statistics method.

### DIFF
--- a/flare/gp_from_aimd.py
+++ b/flare/gp_from_aimd.py
@@ -19,8 +19,8 @@ There are a few reasons why you would want to pay special attention to an
 individual species.
 
 If you are studying a system where the dynamics of one species are
-particularly important and so you want a good representation in the training set,
-then you would want to include as many as possible in the training set
+particularly important and so you want a good representation in the training
+set, then you would want to include as many as possible in the training set
 during the seed part of the training.
 
 Inversely, if a system has high representation of a species well-described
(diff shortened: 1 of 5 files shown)


flare/gp_from_aimd.py
```diff
@@ -182,7 +182,7 @@ def __init__(self, frames: List[Structure],
         assert (isinstance(skip, int) and skip >= 1), "Skip needs to be a " \
                                                       "positive integer."
         self.validate_ratio = validate_ratio
-        assert (validate_ratio >= 0 and validate_ratio <= 1), \
+        assert (0 <= validate_ratio <= 1), \
```
nw13slx, line 185: Didn't know that python can do this. Cool

flare/gp_from_aimd.py
```diff
@@ -243,11 +243,16 @@ def pre_run(self):
                                      self.gp.opt_algorithm,
                                      dt=0,
                                      Nsteps=len(self.frames),
-                                     structure=self.frames[0],
```
nw13slx, line 246: It's good to remove this. It gave me some trouble as well...

tests/test_gp_from_aimd.py
```diff
@@ -255,3 +257,43 @@ def test_mgp_gpfa(all_mgp, all_gp):
...
     tt.run()
+
+
+def test_parse_gpfa_output():
+    """
+    Compare parsing against known answers.
+    :return:
+    """
+    frames, gp_data = parse_trajectory_trainer_output(
+        './test_files/gpfa_parse_test.out', True)
```
nw13slx, line 268: Is it possible to also use the gpfa unit test output as input here? so the test can check whether the code is self consistent?

flare/gp_from_aimd.py
```diff
@@ -559,3 +564,139 @@ def train_gp(self, max_iter: int = None):
...
+
+    # Compute information about GP training
+    # to study GP growth and performance over trajectory
+
+    gp_stats_line = [line for line in lines[:30] if 'GP Statistics' in
+                     line and 'Pre-run' not in line][0][15:].strip()
+
+    initial_gp_statistics = json.loads(gp_stats_line)
+
+    # Get pre_run statistics (if pre-run was done):
```
nw13slx, line 662: do we have a mechanism to check whether the gp model began with some training data before the whole gpfa run? ideally, we can throw a warning to let the user know that the parser will not fully recovered the gp model unless they use the same initial model.

tests/test_gp_from_aimd.py
stevetorr: That's a nice idea and it occurred to me; `gpfa_parse_test.out` is based on a unit test output, but I also wanted to make sure the tests could run consistently. Is there a better way to do this you see?

flare/gp_from_aimd.py
stevetorr: I never know where to set the bar with these things, but, I think it might be okay to trust the user here. The way I see it, the initial GP statistics being empty or not is the signal to the user. However-- I think this is a great idea to keep in our back pocket if/when we write a `rewind GPFA' function, and the warning should be raised then that the GPFA output is not a complete record of the GP's training.

PR #174, opened 2020-05-19
Jon/feature/auto sweep

jonpvandermause:
These changes automatically set the "sweep" value during the construction of atomic environments; this allows the user to choose any cutoff radius, regardless of the size of the structure. This is expected to be useful for small non-cubic structures, where it's not obvious _a priori_ how large the supercell should be when constructing environments.

### DIFF
--- a/flare/env.py
+++ b/flare/env.py
@@ -110,7 +110,7 @@ class AtomicEnvironment:
             cutoffs = dictionary['cutoffs']
         else:
             cutoffs = []
-            for cutoff in ['cutoff_2','cutoff_3','cutoff_mb']:
+            for cutoff in ['cutoff_2', 'cutoff_3', 'cutoff_mb']:
                 if dictionary.get(cutoff):
                     cutoffs.append(dictionary[cutoff])
 
@@ -215,10 +215,12 @@ def get_2_body_arrays(positions: np.ndarray, atom: int, cell: np.ndarray,
 
 
 @njit
-def get_2_body_arrays_ind(positions, atom: int, cell, cutoff_2: float, species):
-    """Returns distances, coordinates, species of atoms, and indexes of neighbors
-    in the 2-body local environment. This method is implemented outside
-    the AtomicEnvironment class to allow for njit acceleration with Numba.
+def get_2_body_arrays_ind(positions, atom: int, cell, cutoff_2: float,
+                          species: np.ndarray):
+    """Returns distances, coordinates, species of atoms, and indexes of
+        neighbors in the 2-body local environment. This method is implemented
+        outside the AtomicEnvironment class to allow for njit acceleration
+        with Numba.
 
     :param positions: Positions of atoms in the structure.
     :type positions: np.ndarray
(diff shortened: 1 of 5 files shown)


flare/struc.py
```diff
@@ -60,6 +60,9 @@ def __init__(self, cell: 'ndarray', species: Union[List[str], List[int]],
         self.vec2 = self.cell[1, :]
         self.vec3 = self.cell[2, :]
 
+        # Compute the max cutoff for sweep = 1.
```
stevetorr, line 63: Without other context, this comment isn't clear ('What is sweep?'). Perhaps something like "Compute the max cell cutoff for later use of periodic images"?

flare/env.py
```diff
@@ -21,12 +21,16 @@ class AtomicEnvironment:
...
         self.structure = structure
         self.positions = structure.wrapped_positions
         self.cell = structure.cell
         self.species = structure.coded_species
-        self.sweep_array = np.arange(-sweep, sweep+1, 1)
+
+        # Set the sweep array based on the 2-body cutoff.
+        sweep_val = ceil(cutoffs[0] / structure.max_cutoff)
+        self.sweep_val = sweep_val
+        self.sweep_array = np.arange(-sweep_val, sweep_val + 1, 1)
```
stevetorr, line 33: Instead of indexing by the 2-body, (which you use because it's assumed to be the largest right?), could also just use the `max` of cutoffs. For instance, we have a 'three body' only kernel, where an unassuming user could (reasonably) think that the 2 body cutoff is a dummy value and set it to a number smaller than it should be, smaller than the 3 body cutoffs, which would be a very subtle, diabolical bug. To my knowledge we have no assert statements which guarantee the 2-body is the biggest.

flare/struc.py
```diff
@@ -127,9 +130,49 @@ def get_cell_dot(self):
 
         return cell_dot
 
+    def get_max_cutoff(self) -> float:
+        """Compute the maximum cutoff compatible with a 3x3x3 supercell of the
```
stevetorr, line 134: This function is nice! Because it takes no arguments and is only used once during the setup of the function, it could be safely moved to a `util` file, and renamed more descriptively via `get_max_cutoff_for_cell` with a more thorough explanation of why it exists (since the util files will be much less frequently scrolled through than `struc.py`).

stevetorr: P.S. It would also just take `cell` as an argument.

Or, if you have strong feelings that ripping it away from the `Structure` class doesn't make sense, it could also be moved away from a method of `Structure` and as a function defined at the bottom of the file.

flare/struc.py
```diff
@@ -433,7 +473,7 @@ def from_file(file_name, format='') -> Union['flare.struc.Structure',
         try:
             with open(file_name, 'r') as _:
                 pass
-        except:
+        except FileNotFoundError:
```
stevetorr, line 476: Nice catch! Literally!

stevetorr:
Penny for your thoughts on the comments above, but otherwise, looks good! Appreciate you keeping everyone Pep8-honest as always.

flare/struc.py
jonpvandermause: Comment clarified, hopefully it makes more sense now.

flare/env.py
jonpvandermause: Excellent catch, I've switched to max.

flare/struc.py
jonpvandermause: Method has been moved to util.py

flare/struc.py
jonpvandermause: This try/except block may be redundant, since I think the same result could be achieved with:
`with open(file_name, 'r') as _:
    pass`

stevetorr: Lol. You're absolutely right.

PR #183, opened 2020-06-09
Add, test method to remove force data from a gp

stevetorr:
In the event of having a model that has bad training data you'd no longer like to use. Removes by internal index of force data; re-computes matrices, updates the internal training label np arrays, and also updates the global data labels. This addition is 100% covered by a unit test which starts with dummy data, makes predictions, adds data, makes new predictions, removes the data, and ensures that the changes this made to the GP were fully reversed.

### DIFF
--- a/flare/gp.py
+++ b/flare/gp.py
@@ -10,7 +10,7 @@ import multiprocessing as mp
 from collections import Counter
 from copy import deepcopy
 from numpy.random import random
-from typing import List, Callable, Union
+from typing import List, Callable, Union, Tuple
 from scipy.linalg import solve_triangular
 from scipy.optimize import minimize
 
@@ -835,7 +835,8 @@ class GaussianProcess:
             self.train()
 
     def remove_force_data(self, indexes: Union[int, List[int]],
-                          update_matrices: bool = True):
+                          update_matrices: bool = True)->Tuple[List[Structure],
+                                                              List['ndarray']]:
         """
         Remove force components from the model. Convenience function which
         deletes individual data points.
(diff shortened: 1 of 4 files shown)


flare/gp.py
```diff
@@ -823,7 +834,48 @@ def adjust_cutoffs(self, new_cutoffs: Union[list, tuple, 'np.ndarray'],
...
+        # Listify input even if one integer
+        if isinstance(indexes, int):
+            indexes = [indexes]
+
+        if max(indexes) > len(self.training_data):
+            raise ValueError("Index out of range of data")
+
+        # Get in reverse order so that modifying higher indexes doesn't affect
+        # lower indexes
+        indexes.sort(reverse=True)
```
nw13slx, line 866: That's clever.

flare/gp.py
```diff
@@ -823,7 +834,48 @@ def adjust_cutoffs(self, new_cutoffs: Union[list, tuple, 'np.ndarray'],
...
 
+        for i in indexes:
+            del self.training_data[i]
+            del self.training_labels[i]
+
+        self.training_labels_np = np.hstack(self.training_labels)
+        _global_training_data[self.name] = self.training_data
+        _global_training_labels[self.name] = self.training_labels_np
```
nw13slx, line 874: The self.all_labels may also need to be updated.

flare/gp.py
```diff
@@ -823,7 +834,48 @@ def adjust_cutoffs(self, new_cutoffs: Union[list, tuple, 'np.ndarray'],
...
+
+        if update_matrices:
+            self.set_L_alpha()
+            self.compute_matrices()
```
nw13slx, line 879: maybe consider returning the deleted data and labels like a pop function.

PR #184, opened 2020-06-11
Better user interface and consistency check through out env/gp/kernel modules

nw13slx:
Major changes:

1. Input arguments:
- cutoff is no longer an array. but a dictionary   {'twobody': 7, 'threebody':4.5}
- kernel_name, kernel functions are all removed. replace by "kernels" that takes an array of all different kernels ['twobody', 'threebody', 'manybody']
2. new class:
- Parameters class and methods that check whether hyper-parameters, hyper-parameter-mask, cutoffs, kernels are consistent. All relevant checks and info retrival in GP, kernel/utils, mgp are shifted to the parameters.py . (hopefully it's easier to maintain.)
3. special kernel for MGP that loops over grid point, xyz, and triplets inside the kernel.
4. all prints are replaced by logger. Output class is rewritten with logger. So it can be easier to print to screen and files at the same time. As a result, the verbose level of GPFA and OTF are all changed from "1-5" to "error, warning, info, debug"

Minor changes:
1. backward_arguments and backward_attributes are added to GP. All old changes in argument and attributes over different versions of Flare are documented there. So old json and pickle file can still be loaded.
2. a couple more functions are added to GP class to reduce repeated lines
3. OTF and GPFA header printing, parser, GP reconstruction are slightly revised to make it easier to maintain.
4. unit tests were adapted

An example of the new GP initialization:

```python
from flare.gp import GaussianProcess
from flare.utils.parameter_helper import ParameterHelper

pm = ParameterHelper(species=['Cu', 'C', 'H', 'O'],
                          kernels={'twobody':[['*', '*'], ['Cu','Cu']],
                                  'threebody':[['*', '*', '*'], ['Cu','Cu', 'Cu']]},
                           parameters={'twobody0':[1.29, 1.47], 'twobody1':[7e-3, 0.19],
                                 'threebody0':[8e-4, 0.89], 'threebody1':[6.1e-3, 0.72],
                                 'cutoff_twobody':7.0, 'cutoff_threebody':4.5,
                                 'noise':0.05, 'energy_noise':0.01})
hm = pm.as_dict()
cutoffs = hm['cutoffs']
hyps = hm['hyps']
hyp_labels = hm['hyp_labels']
kernels = hm['kernels']
 
gp_model = GaussianProcess(kernels=kernels,
                                                 hyps=hyps, cutoffs=cutoffs,
                                                 hyp_labels=hyp_labels,
                                                 hyps_mask=hm)
```


YuuuuXie:
The MGP is done. @jonpvandermause @stevetorr @nw13slx

YuuuuXie:
The ASE interface changes include:
1. Previously, we have `flare/ase/otf.py` which is independent of the `flare/otf.py`. Now I removed the original one, and have `flare/ase/otf.py` inherit from `flare/otf.py`
2. Previously, we also have `flare/ase/logger.py` which tries to reproduce the log file output of `flare/otf.py`. Since now we have the inheritance as described in 1), this is no longer needed, thus deleted
3. Removed `ase/otf_md.py`. Previously, the ASE OTF interface inherit ASE_OTF module and MD module of ASE. Now the various md engines (VelocityVerlet, Langevin etc.) are used as an input of the otf training. See `test_ase_otf.py`, or the tutorial `ase.ipynb`

docs/source/tutorials/prepare_data.rst
```diff
@@ -0,0 +1,67 @@
...
+To have a more complete and better monitored training process, please use our 
+`GPFA module <https://flare.readthedocs.io/en/latest/tutorials/gpfa.html>`_. 
+
+Here we are not going to use this module, but only provide a simple example on 
+how the GP is constructed from the data.
+
+.. code-block:: python
+
+    from flare.gp import GaussianProcess
+    from flare.parameters import Parameters
```
nw13slx, line 67: unfinished?

flare/dft_interface/vasp_util.py
```diff
@@ -86,9 +88,10 @@ def run_dft(calc_dir: str, dft_loc: str,
         try:
             forces = parse_func("vasprun.xml")
         except FileNotFoundError:
-            raise FileNotFoundError("""Could not load vasprun.xml.
-            The calculation may not have finished.
-            Current directory is %s""" % os.getcwd())
+            os.chdir(currdir)
```
nw13slx, line 91: It's great to add chdir before the raise!

flare/mgp/mapxb.py
```diff
@@ -0,0 +1,655 @@
...
+        raise NotImplementedError("need to be implemented in child class")
+
+    def build_map_container(self, bounds):
+        '''
+        construct an empty spline container without coefficients.
+        '''
+
+        self.maps = []
+        for spc in self.spc:
+            m = self.singlexbody((self.grid_num, bounds, spc,
```
nw13slx, line 80: @YuuuuXie  maybe we should explicitly write out which variable is used for which argument. For example (grid_num=self.grid_num, bounds=bounds, ... ). So it's easier to debug if we change the init arguments of singlexbody in the future.

flare/mgp/mapxb.py
```diff
@@ -0,0 +1,655 @@
...
+
+        min_dist = atom_env.bond_array_2[0][0]
+        lower_bound = np.max(self.maps[0].bounds[0][0])
+        assert min_dist >= lower_bound,\
+                f'The minimal distance {min_dist:.3f} is below the mgp lower bound {lower_bound:.3f}'
+
+        if self.mean_only:  # if not build mapping for var
+            mean_only = True
+
+        force_kernel, en_kernel, _, cutoffs, hyps, hyps_mask = self.kernel_info
```
nw13slx, line 108: should we double check whether hyps_mask is consistent with the cutoffs_mask in atom_env ?

flare/mgp/mapxb.py
```diff
@@ -0,0 +1,655 @@
...
+
+    def get_arrays(self, atom_env):
+        raise NotImplementedError("need to be implemented in child class")
+
+    def build_map_container(self, bounds):
+        '''
+        construct an empty spline container without coefficients.
+        '''
+
+        self.maps = []
```
nw13slx, line 78: self.maps = [] should also be decalred in the init function

flare/mgp/mapxb.py
```diff
@@ -0,0 +1,655 @@
...
+
+        if 'container_only' not in dictionary:
+            dictionary['container_only'] = True
+
+        # initialize
+        init_args_name = ['grid_num', 'lower_bound', 'upper_bound', 'svd_rank',
+            'species_list', 'map_force', 'GP', 'mean_only', 'container_only', 
+            'lmp_file_name', 'load_grid', 'lower_bound_relax', 'n_cpus', 'n_sample']
+        args = [dictionary[name] for name in init_args_name]
+        new_mgp = mapxbody(args)
```
nw13slx, line 190: I think the above block could be replace by 
```python
new_mgp = mapxbody(**args)
```
So you don't need to maintain two argument lists (here and the init functions)

flare/mgp/mapxb.py
```diff
@@ -0,0 +1,655 @@
...
+        """
+
+        out_dict = deepcopy(dict(vars(self)))
+
+        # Uncertainty mappings currently not serializable;
+        if not self.mean_only:
+            out_dict['mean_only'] = True
+
+        # save the kernel name instead of callables
+        kernel, ek, efk, cutoffs, hyps, hyps_mask = self.kernel_info
```
nw13slx, line 154: this block of code is the same as line 108. It should be isolated out as a function. And add a "kernel_info" argument or something equivalent to the init function

flare/mgp/mapxb.py
```diff
@@ -0,0 +1,655 @@
...
+            'species_list', 'map_force', 'GP', 'mean_only', 'container_only', 
+            'lmp_file_name', 'load_grid', 'lower_bound_relax', 'n_cpus', 'n_sample']
+        args = [dictionary[name] for name in init_args_name]
+        new_mgp = mapxbody(args)
+
+        # Restore kernel_info
+        kernel_info = dictionary['kernel_info']
+        kernel_name = kernel_info[0]
+        hyps_mask = kernel_info[-1]
+        kernel, _, ek, efk = str_to_kernel_set([kernel_name], 'mc', hyps_mask)
```
nw13slx, line 196: this block of code is the same as line 108. It should be isolated out as a function. And please consider adding a "kernel_info" argument (or something equivalent) to the init function

flare/output.py
```diff
@@ -50,59 +51,55 @@ def conclude_run(self):
...
-        del self.outfiles
-        self.outfiles = {}
+        logger = logging.getLogger(self.basename+'log')
+        logger.info('-' * 20)
+        logger.info('Run complete.')
+        logging.shutdown()
+        self.logger = []
 
-    def open_new_log(self, filetype: str, suffix: str):
+    def open_new_log(self, filetype: str, suffix: str, verbose='info'):
```
jonpvandermause, line 60: Can we set the default verbosity to a lower level, e.g. 'warning' or 'error'? Otherwise almost everything in an otf run gets printed to the console.

nw13slx: the screen output should be disable by the latest commit

flare/otf.py
```diff
@@ -191,69 +202,50 @@ def run(self):
...
                                     self.gp.hyps[-1], self.structure,
                                     self.max_atoms_added)
 
-                if not std_in_bound:
+                if std_in_bound:
+                    new_pos = self.md_step()
+
+                else:
                     # record GP forces
                     self.update_temperature(new_pos)
```
jonpvandermause, line 229: The logic is wrong here; velocities get sent to zero, since new_pos hasn't been updated. (I'll fix this)

flare/ase/otf.py
```diff
@@ -1,358 +1,182 @@
...
-        atom_count = 0
-        atom_list = []
-        gp_model = self.atoms.calc.gp_model
+    def md_step(self): 
+        '''
+        Get new position in molecular dynamics based on the forces predicted by
+        FLARE_Calculator or DFT calculator
+        '''
+        self.md.step()
+        return self.atoms.positions
```
jonpvandermause, line 158: This is different from the md_step method in otf.py, which computes the new positions but doesn't overwrite the positions attribute of the structure

flare/ase/otf.py
```diff
@@ -1,358 +1,182 @@
...
-    def update_GP(self, dft_forces):
-        atom_count = 0
-        atom_list = []
-        gp_model = self.atoms.calc.gp_model
+    def md_step(self): 
+        '''
+        Get new position in molecular dynamics based on the forces predicted by
+        FLARE_Calculator or DFT calculator
+        '''
+        self.md.step()
```
jonpvandermause, line 157: Somehow, self.md.step() updates the positions of self.structure, even though it should act only on self.atoms. This causes the velocities and temperatures to be reported as zero in the OTF output file.

flare/ase/otf.py
jonpvandermause: I've attempted to resolve this with 125c08f by removing the second call to md.step() in the otf loop

flare/ase/otf.py
jonpvandermause: 1829b0c seems to give a temporary fix to this problem by manually resetting the positions of self.structure. But I'm still confused why self.md.step() updates self.structure.positions...

flare/otf.py
jonpvandermause: 8544b9f addresses this by mirroring the otf loop on master (new positions are calculated first, and then is_std_in_bound is called)

flare/ase/otf.py
jonpvandermause: The issue was that self.md.step() returned atoms.positions rather than a copy of it, which caused self.atoms and self.structure to get updated together downstream. a2a785a fixes this

flare/gp.py
```diff
@@ -34,18 +34,14 @@ class GaussianProcess:
     Williams.
 
     Args:
-        kernel (Callable, optional): Name of the kernel to use, or the kernel
-            itself.
-        kernel_grad (Callable, optional): Function that returns the gradient
-            of the GP kernel with respect to the hyperparameters.
+        kernels (list, optional): Determine the type of kernels. Example:
+            ['2', '3'], ['2', '3', 'mb'], ['2']. Defaults to ['2', '3']
```
stevetorr, line 38: Nice change.

flare/gp.py
```diff
@@ -682,33 +625,29 @@ def as_dict(self):
 
         return out_dict
 
+    def sync_data(self):
```
stevetorr, line 628: Maybe we should move the setting of the NP training labels to this function, so that it's only a matter of adding to training data and training labels?

stevetorr: Updating the 'all labels' attribute could also happen here.

flare/gp.py
```diff
@@ -174,82 +162,58 @@ def check_instantiation(self):
...
-        else:
-            self.multihyps = False
-            self.hyps_mask = None
+        self.bounds = deepcopy(self.hyps_mask.get('bounds', None))
 
-    def update_kernel(self, kernel_name, multihyps=False):
-        kernel, grad, ek, efk = str_to_kernel_set(kernel_name, multihyps)
+    def update_kernel(self, kernels, component="mc", hyps_mask=None):
+        kernel, grad, ek, efk = str_to_kernel_set(
+            kernels, component, hyps_mask)
```
stevetorr, line 211: No docstring? :)

flare/gp.py
```diff
@@ -962,3 +933,76 @@ def __del__(self):
         if (self.name in _global_training_labels):
             _global_training_labels.pop(self.name, None)
             _global_training_data.pop(self.name, None)
+
+    @staticmethod
+    def backward_arguments(kwargs, new_args={}):
```
stevetorr, line 938: Love these deprecation warnings. It's great we have this! Will save a lot of grief in using old scripts

flare/gp_from_aimd.py
```diff
@@ -74,7 +75,7 @@ def __init__(self, frames: List[Structure],
                  max_trains: int = np.inf,
                  min_atoms_per_train: int = 1,
                  shuffle_frames: bool = False,
-                 verbose: int = 1,
+                 verbose: str = "INFO",
```
stevetorr, line 78: Great call to re-do it this way

flare/kernels/kernels.py
```diff
@@ -36,15 +36,21 @@ def force_helper(A, B, C, D, fi, fj, fdi, fdj, ls1, ls2, ls3, sig2):
...
+
+#    E = exp(-D * ls1)
+#    F = E * B * ls2
+#    G = -E * C * ls2
+#    H = A * E * ls2 - B * C * E * ls3
+#    I = E * fdi * fdj
+#    J = F * fi * fdj
+#    K = G * fdi * fj
+#    L = H * fi * fj
+#    M = sig2 * (I + J + K + L)
```
stevetorr, line 53: Any reason why these are hanging around?

flare/mgp/utils.py
```diff
@@ -114,27 +120,36 @@ def get_triplets(ctype, etypes, bond_array, cross_bond_inds,
...
+#            coord = [c1, c2] 
+#
+#            if spcs not in exist_species:
+#                exist_species.append(spcs)
+#                tris += [triplet]
+#                tri_dir += [coord]
+#            else:
+#                k = exist_species.index(spcs)
+#                tris[k] += triplet
+#                tri_dir[k] += coord
```
stevetorr, line 146: Any reason why these guys are still here?

flare/kernels/kernels.py
YuuuuXie: yeah we can remove those commented lines, they were for debugging

tests/test_otf.py
```diff
@@ -13,12 +14,16 @@
 example_list = [1, 2]
 name_list = {1:'h2', 2:'al'}
 
+print('running test_otf.py')
+print('current working directory:')
+print(os.getcwd())
```
stevetorr, line 19: I suspect these were accidentally left over from debugging?

tests/test_otf.py
```diff
@@ -55,6 +60,10 @@ def test_otf(software, example):
     :return:
     """
 
+    print('running test_otf.py')
+    print('current working directory:')
+    print(os.getcwd())
```
stevetorr, line 65: Likewise :)

stevetorr:
LGTM (looks good to me). **_Major_** kudos to @nw13slx  and @YuuuuXie for all their blood, sweat, and tears in making FLARE easier to use and maintain.

PR #185, opened 2020-06-12
Revert "Add, test method to remove force data from a gp" due to mysterious unit test failure

stevetorr:
Reverts mir-group/flare#183


jonpvandermause:
This is probably an issue with the CP2K executable, not your commit (sometimes the site is down). Shouldn't be necessary to revert

stevetorr:
Problem lied elsewhere; solved

PR #186, opened 2020-06-12
Attempt to solve unit tests failing due to pytest

stevetorr:
Travis logs indicate that pytest is at too low a version (perhaps due to a change in some requirement by another library). I have fixed this by adding a minimum version to pytest in the requirements.txt.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ numba
 ase
 pymatgen
 nptyping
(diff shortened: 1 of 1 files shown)


jonpvandermause:
Good catch, this seems to do the trick

stevetorr:
Thanks! I'll pull this once the travis passes.

PR #187, opened 2020-06-18
Stress kernels and other kernel related changes

jonpvandermause:
This is a step toward NPT on-the-fly training with the GP. Summary of major changes/additions:

**mc_simple**
- 2, 3, and 2+3-body stress kernels (stress/energy, stress/force, stress/stress)
- Simultaneous calculation of energy, force and stress kernels to accelerate prediction (i.e. no d1, d2 arguments; see efs_energy and efs_force functions)
- "efs self" kernel to facilitate the calculation of uncertainties on energies and stresses
- To do (in a future PR): implement efs kernels in mc_sephyps, mc_mb_sepcut; eliminate d1, d2 arguments from all kernels

**test_stress_kernels**
- Careful test of all mc_simple energy, force, and stress kernels, including tests of the fractional pre-factors.
- TODO: apply this test to all existing kernels

**two_body_mc_simple, three_body_mc_simple**
- This is a first draft for the future kernel class. It hasn't been integrated into the rest of the code base yet.

**predict**
- Simultaneous energy, force and stress prediction (serial and parallel) without redundant kernel calls (see predict_on_structure_efs)
- GP uncertainties on energies and stresses
- To do: accelerate predict_on_structure and predict_on_structure_en

**gp_algebra**
- efs_kern_vec: computes kernel vectors for energy, force, and stress prediction without redundant kernel calls

**ase/calculator**
- get_stress method added to the gp calculator
- DFT stresses and energy are now written to structure 

**ase/otf**
- Velocities, temperature and KE are now based on ASE atoms attributes
- Tighter coupling between atoms and structure
- Units corrections (time now reported in ps in the output file)

**output**
- Periodic cell, stress tensor (with uncertainties), and pressure are now reported in the OTF output file

**struc**
- positions, cell, and related arrays are now properties. When the positions or cell attributes are reset, all the related arrays (e.g. wrapped positions, cell transpose, etc.) are automatically changed.


flare/kernels/three_body_mc_simple.py
```diff
@@ -0,0 +1,1263 @@
...
+                    r13 = ri1 - rj3
+                    r21 = ri2 - rj1
+                    r22 = ri2 - rj2
+                    r23 = ri2 - rj3
+                    r31 = ri3 - rj1
+                    r32 = ri3 - rj2
+                    r33 = ri3 - rj3
+
+                    stress_count = 0
+                    for d1 in range(3):
```
nw13slx, line 422: data locality would be better if the d1 and d2 loop is moved to the right before the helper function and let ci1, ci2 to be 1d vector. so fi and fj won't be computed for 9 times.

jonpvandermause: Addressed in 6952612

flare/ase/otf.py
```diff
@@ -164,8 +170,17 @@ def update_positions(self, new_pos):
...
+    def update_temperature(self):
+        self.KE = self.atoms.get_kinetic_energy()
+        self.temperature = self.atoms.get_temperature()
+
+        # Convert velocities to Angstrom / ps.
+        self.velocities = self.atoms.get_velocities() * units.fs * 1e3
 
+    def run_dft(self):
+        super().run_dft()
+
```
YuuuuXie, line 182: We can remove this `run_dft` function, if it is not defined, it will automatically use the `run_dft` function in the parent class

flare/kernels/two_body_mc_simple.py
```diff
@@ -0,0 +1,712 @@
...
+                          env2.bond_array_2, env2.ctype, env2.etypes,
+                          self.signal_variance, self.length_scale,
+                          self.cutoff, self.cutoff_func)
+
+    def efs_force(self, env1: AtomicEnvironment, env2: AtomicEnvironment):
+        return efs_force(env1.bond_array_2, env1.ctype, env1.etypes,
+                         env2.bond_array_2, env2.ctype, env2.etypes,
+                         self.signal_variance, self.length_scale,
+                         self.cutoff, self.cutoff_func)
+
```
YuuuuXie, line 78: Since all the inputs are the same, perhaps we can define a function like:
```
def get_args(self, env1, env2):
    return (env1.bond_array_2, env1.ctype, env1.etypes,
                          env2.bond_array_2, env2.ctype, env2.etypes,
                          self.signal_variance, self.length_scale,
                          self.cutoff, self.cutoff_func)
```

and then we do 
```
def efs_force(self, env1: AtomicEnvironment, env2: AtomicEnvironment):
    args = self.get_args(env1, env2)
    return efs_force(*args)
```
same for energy_energy, force_force etc.

nw13slx: and maybe this get args can be use along with the mask_to_args. would be something like
```
hyp_args = mask_to_args(hyps, hyps_mask)
env_args = get(args, env1, env2)
args = env_args + hyp_args
```

flare/ase/calculator.py
```diff
@@ -11,139 +11,129 @@
...
         self.results = {}
 
-    def get_property(self, name, atoms=None, allow_calculation=True):
+    def get_property(self, name, atoms=None, allow_calculation=True,
+                     structure=None):
         if name not in self.results.keys():
             if not allow_calculation:
                 return None
-            self.calculate(atoms)
+            self.calculate(atoms, structure)
```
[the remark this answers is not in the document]
jonpvandermause: The interface as implemented is consistent with the general ASE calculator template described [here](https://wiki.fysik.dtu.dk/ase/ase/calculators/calculators.html#calculator-interface). Note that our get_forces, get_potential_energy, and get_stress methods take in an atoms object only; the structure object is only introduced as an optional input to the auxiliary get_property method.

flare/kernels/kernels.py
```diff
@@ -110,6 +108,90 @@ def three_body_helper_2(ci1, ci2, cj1, cj2, r12, r23, r31,
...
+    D = r11 * r11 + r22 * r22 + r33 * r33
+    E = exp(-D * ls1)
+    F = E * B * ls2
+    G = -E * C * ls2
+    H = A * E * ls2 - B * C * E * ls3
+    I = E * (fdi1 * coord1 + fdi2 * coord2) * fdj
+    J = F * fi * fdj
+    K = G * (fdi1 * coord1 + fdi2 * coord2) * fj
+    L = H * fi * fj
+    M = -sig2 * (I + J + K + L)
```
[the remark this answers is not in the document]
jonpvandermause: Good catch, I've tried to address them in 095eba9.

flare/kernels/kernels.py
```diff
@@ -145,6 +227,323 @@ def three_body_en_helper(ci1, ci2, r11, r22, r33, fi, fj, fdi, ls1, ls2, sig2):
...
+            C2 = r11 * r11 + r23 * r23 + r32 * r32
+            kern += sig2 * exp(-C2 * ls2) * fi * fj
+    if (c1 == ej2):
+        if (ei1 == ej1) and (ei2 == c2):
+            C6 = r13 * r13 + r22 * r22 + r31 * r31
+            kern += sig2 * exp(-C6 * ls2) * fi * fj
+        if (ei1 == c2) and (ei2 == ej1):
+            C4 = r12 * r12 + r23 * r23 + r31 * r31
+            kern += sig2 * exp(-C4 * ls2) * fi * fj
+
```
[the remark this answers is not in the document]
jonpvandermause: Another good catch, I've addressed this in 5f03674. (Let me know if you see other redundancies)

flare/otf.py
```diff
@@ -87,28 +89,25 @@ class OTF:
...
+     std_tolerance_factor: float = 1, skip: int = 0,
+     init_atoms: List[int] = None, output_name: str = 'otf_run',
+     max_atoms_added: int = 1, freeze_hyps: int = 10,
+     # dft args
+     force_source: str = "qe", npool: int = None, mpi: str = "srun",
+     dft_loc: str = None, dft_input: str = None, dft_output='dft.out',
+     dft_kwargs=None,
+     store_dft_output: Tuple[Union[str, List[str]], str] = None,
+     # par args
+     n_cpus: int = 1):
```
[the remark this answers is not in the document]
jonpvandermause: This is a very good idea, and I think Black would be a good choice. Let's discuss this at the next dev meeting

flare/ase/otf.py
jonpvandermause: Thanks for catching this, I've removed the method in 9b13052

flare/ase/calculator.py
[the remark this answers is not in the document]
YuuuuXie: Actually all the `get_forces`, `get_stress` etc. calls `get_property`, the reason `atoms.get_forces()` works in `otf` is that the line 144 (ase/otf.py)
```
        self.atoms.calc.calculate(self.atoms, self.structure)
```
called the `calculate` function, so there is a dict `self.atoms.calc.results` which has been written the FLARE calculated `forces`, `energy` etc. Then when the `atoms.get_forces` is called, it will just get the value from the `results` dict without re-calculating of GP, so it's ok at this time if `structure=None`.

This means that it only works in the `otf` context. If we want to create a FLARE_Calculator without OTF, and try to use `get_forces`, the structure has to be input, because at this time `results` dict is empty, and `calculate` function needs calling.

But I notice that `structure` and ASE Atoms are very similar, so probably a better way to do this is just in `ASE_OTF` let
```
    self.structure = self.atoms
```
since they have a lot of common attributes like positions, cell etc. so there should only be some minor changes. may have to think about this. I will try to deal with this in another PR

flare/kernels/two_body_mc_simple.py
jonpvandermause: Good idea, this cuts down on a lot of unnecessary lines. I added get_args methods to the 2- and 3-body kernels in 5633d98.

flare/ase/calculator.py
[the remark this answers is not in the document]
jonpvandermause: Hmm I actually think get_forces, get_stress etc. should still work without giving a structure. If a structure isn't given, the calculate method creates a structure based on the atoms object (see lines 69-73 of ase/calculator). But I agree this isn't a perfect solution, and it would be better to completely eliminate structure from the ase calculator. This would require adding some extra calculator properties (energy, force, stress uncertainties) and then writing them to the structure somewhere else, probably in a method of ase/otf.

PR #188, opened 2020-06-18
Steven/feature/gp write choice

stevetorr:
Minor changes: 
- `GaussianProcess.write_model` now allows for flexibility for the user to decide if they want the matrices written separately or not. 
- `write_model` now will not redundantly append .json or .pickle to written models. If the file name is `my_model.json` then it will not write `my_model.json.json`. Also, if you don't specify the format, but if you specify it in the string, it automatically detects it. A small friction I've encountered!
- The remove force data method will now terminate immediately if an empty list is passed in.

### DIFF
--- a/flare/gp.py
+++ b/flare/gp.py
@@ -797,6 +797,9 @@ class GaussianProcess:
         if max(indexes) > len(self.training_data):
             raise ValueError("Index out of range of data")
 
+        if len(indexes) == 0:
+            return [], []
+
         # Get in reverse order so that modifying higher indexes doesn't affect
         # lower indexes
         indexes.sort(reverse=True)
--- a/flare/gp.py
+++ b/flare/gp.py
@@ -705,14 +705,21 @@ class GaussianProcess:
         :return:
         """
         ky_mat = self.ky_mat
-        l_mat = np.linalg.cholesky(ky_mat)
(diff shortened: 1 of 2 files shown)


flare/gp.py
```diff
@@ -670,20 +673,24 @@ def from_dict(dictionary):
         new_gp.n_envs_prev = len(new_gp.training_data)
 
         # Save time by attempting to load in computed attributes
-        if len(new_gp.training_data) > 5000:
```
nw13slx, line 673: thanks for removing this arbitrary criteria

flare/gp.py
```diff
@@ -670,20 +673,24 @@ def from_dict(dictionary):
...
+        if dictionary.get('ky_mat_file'):
```
nw13slx, line 676: maybe next time this part can be written as a function? it repeats again in the from_dict method.

flare/gp.py
```diff
@@ -829,21 +855,35 @@ def write_model(self, name: str, format: str = 'json'):
             self.alpha = None
             self.ky_mat_inv = None
 
+        # Automatically detect output format from name variable
+
+        for detect in ['json','pickle','binary']:
```
nw13slx, line 860: this is very thoughtful

PR #191, opened 2020-06-23
fix optimization flag in ParameterHelper class

nw13slx:
ParameterHelper failed to apply the right constraints defined by set_constraints() functions, because the order was messed up... 
If user use ParameterHelper to generate hyps_mask for constrained hyper-parameter optimization, it will optimized the wrong element of the hyper-parameters...


flare/utils/parameter_helper.py
```diff
@@ -844,10 +845,11 @@ def summarize_group(self, group_type):
...
                     self.hyps_sig[group_type] += [sig]
                     self.hyps_ls[group_type] += [ls]
-                    self.hyps_opt[group_type] += [opt_sig]
-                    self.hyps_opt[group_type] += [opt_ls]
+                    all_opt_sig += [opt_sig]
+                    all_opt_ls += [opt_ls]
                     self.logger.debug(f"   using hyper-parameters of {sig:6.2g} "
-                                      f"{ls:6.2g}")
+                                      f"{ls:6.2g} {opt_sig} {opt_ls}")
+            self.hyps_opt[group_type] = all_opt_sig + all_opt_ls
```
stevetorr, line 852: So to confirm, it's okay that it's all signal hyps and _then_ all length scale params within the hyps_opt indexes? If so, this looks good to me.

stevetorr:
At the risk of making a trite critique: Is it possible to unit test this logic (that constraints on optimization operate in the right way) slash were the previous unit tests not robust enough to catch this?

flare/mgp/mgp.py
```diff
@@ -111,13 +111,21 @@ def __init__(self,
...
-                self.coded_species.append(ele)
-                self.species_labels.append(Z_to_element(ele))
+                coded_species.append(ele)
+                species_labels.append(Z_to_element(ele))
+            else:
+                print("element type not accepted", ele, type(ele))
+        sort_id = np.argsort(coded_species)
+        for i in sort_id:
+            self.coded_species.append(coded_species[i])
+            self.species_labels.append(species_labels[i])
```
stevetorr, line 128: Nice!

stevetorr: A quick question: Is there anything that we could do here to make this future-proofed for coarse graining? Maybe it's thinking too far ahead, but just so that if a user were to pass in a number outside of {1...118} or a species label outside of the periodic table that everything ends up working okay. I think that this might be too far in advance (and the logic here is easy enough that we could fix it later) but something to keep in mind. This could also matter sooner if someone wants to do FLARE with isotopes, to keep track of distinct species.

flare/utils/parameter_helper.py
nw13slx: that's exactly where the bug is. The hyper-parameter array is always ordered like "sig2-1 sig2-2 ls2-1 ls2-2 sig3-1 sig3-2 ..."

flare/mgp/mgp.py
nw13slx: I guess we can change the element_to_Z part and allow hyps_mask to take any number beyond 118. I can change the ParameterHelper to accommodate that.

But we need to think a bit how to make it consistent. Maybe leave a dictionary in the element_coder module that allows users to injet str-integer pair. And the element_to_Z and Z_to_element function can read from that dictionary.

stevetorr:
Thanks for adding those tests!! (a favor not only to the future version of you but also the future developers and contributors... :) )

PR #195, opened 2020-06-29
Feature/lixin/gpfa improvement

nw13slx:
1. gp_from_aimd: 
    - print potential energy for mgp
    - allow printing dft configurations in xyz format
2. element_coder:
    - allow user to define their own element species
3. gp
    - while training hyper-parameters, pass the name of the logger as argument, instead of the Logger object
4. bugfix for constrained optimization in gpfa and otf
    - the original version in development branch does not print the right hyper-parameters if 'map' flag is used for constraint optimization
    - it also reads the wrong noise parameter


flare/output.py
```diff
@@ -355,7 +359,8 @@ def write_xyz_config(self, curr_step, structure, dft_step,
         self.write_xyz(curr_step=curr_step, pos=structure.positions,
                        species=structure.species_labels, filename='xyz',
                        header=header,
-                       forces=forces, stds=stds, forces_2=forces_2)
+                       forces=forces, stds=stds, forces_2=forces_2,
```
stevetorr, line 362: I implemented a `print_as_xyz` method in the `Structure` class which is fairly flexibile; perhaps this could be used here?

flare/output.py
```diff
@@ -41,19 +41,23 @@ class Output:
 
     def __init__(self, basename: str = 'otf_run',
                  verbose: str = 'INFO',
+                 print_xyz: bool = False,
                  always_flush: bool = False):
         """
         Construction. Open files.
         """
         self.basename = f"{basename}"
+        self.print_xyz = print_xyz
```
stevetorr, line 50: If you agree with the `print_as_xyz` change in `gp_from_aimd.py` then we can refactor it here. I think that it makes it slightly more clear that it is changing the format of the printing to call it print *as* xyz, instead of printing xyz format in addition to what is ordinarily printed.

flare/output.py
```diff
@@ -407,15 +412,16 @@ def add_atom_info(self, train_atoms, stds):
         f.info(f'Uncertainty: {stds[train_atoms[0]]}')
 
     def write_gp_dft_comparison(
-     self, curr_step, frame, start_time, dft_forces, error,
-     local_energies=None, KE=None, mgp=False):
+            self, curr_step, frame, start_time, dft_forces, dft_energy, error,
+            local_energies=None, KE=None, mgp=False):
```
stevetorr, line 416: Maybe make DFT energy optional, since that label might not be included?

flare/output.py
```diff
@@ -488,13 +494,20 @@ def write_gp_dft_comparison(
             stat += f' {mae_per_species[ele]:.2f}'
 
         # calculate potential and total energy
+        pot_en = 0
```
stevetorr, line 497: Maybe tuck this into the if statement below?

flare/parameters.py
```diff
@@ -230,8 +230,10 @@ def check_instantiation(hyps, cutoffs, kernels, param_dict):
                             f'number of cutoffs should be the same as n {n}'
                         assert npmax(cutoff_list) <= cutoffs[kernel]
             else:
-                assert f'{kernel}_mask' not in param_dict
-                assert f'{kernel}_cutof_list' not in param_dict
+                assert f'{kernel}_mask' not in param_dict,\
+                        f'{kernel}_mask should not be in param_dict'
+                assert f'{kernel}_cutoff_list' not in param_dict, \
+                        f'{kernel}_cutoff_list should not be in param_dict'
```
stevetorr, line 236: Always good to make it clearer to the user!

flare/parameters.py
```diff
@@ -421,7 +423,7 @@ def get_cutoff(kernel_name, coded_species, param_dict):
                 return [universal_cutoff]*3
 
     @staticmethod
-    def get_hyps(param_dict, hyps=None, constraint=False):
+    def get_hyps(param_dict, hyps=None, constraint=False, label=False):
```
stevetorr, line 426: Docstring needs to be updated with what label does

stevetorr: (and just updated period)

flare/utils/element_coder.py
```diff
@@ -8,6 +8,36 @@
...
+    elements in the system
+
+    :param element: string symbol of the element
+    :type element: str
+    :param Z: corresponding Z
+    :type Z: int
+    """
+    _user_element_to_Z[element] = Z
+    _user_Z_to_element[Z] = element
+
```
stevetorr, line 40: Nice future proofing. The coarse-grain / isotopic users of FLARE will be grateful for this one day.

flare/output.py
nw13slx: cool. changed in e235d15a17aeda3065bb27a9d290dc06209a9ff0

flare/output.py
nw13slx: set the default to NaN in commit 485810945624f83687b50f0dc76d529dbc39eb34

PR #196, opened 2020-07-02
Implement structure in C++

jonpvandermause:
Note: I've included all the flare++ code in this PR, but I've only integrated the structure class with the rest of the repo. In the next week or so, I'll be integrating env and kernels.

**Sources/Structure**
- directory containing the C++ implementation of the structure class
- py_structure.h and py_structure.cpp create the pybind11 binding

**structure.py**
- keeps quality-of-life methods (as_dict, str, len, etc.) in pure Python

**env.py**
- as a first step toward a C++ implementation of the local environment class, the cutoffs mask has been changed from a dictionary to a class.

**Sources/Parameters**
- Bare bones C++ implementation of the hyperparameter mask.

**setup.py**
- pip now compiles the C++ code and downloads all dependencies (googletest, Eigen, and pybind11) from github


CMakeLists.txt
```diff
@@ -0,0 +1,149 @@
+cmake_minimum_required(VERSION 3.12)
+
+set(CMAKE_DISABLE_SOURCE_CHANGES ON)
+set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
+
+project(flare LANGUAGES C CXX)
```
anjohan, line 6: is there pure C code anywhere?

CMakeLists.txt
```diff
@@ -0,0 +1,149 @@
...
+
+
+#
+# Dependencies
+#
+
+# googletest
+################################################################################
+
+# Download and unpack googletest at configure time
```
anjohan, line 32: I think this can be handled by externalProject, I think.

Or
```
FetchContent_Declare(googletest
    GIT_REPOSITORY https://github.com/google/googletest.git
    SOURCE_DIR ${CMAKE_CURRENT_BINARY_DIR}/googletest
    UPDATE_COMMAND ""
)
FetchContent_MakeAvailable(googletest)
```

CMakeLists.txt
```diff
@@ -0,0 +1,149 @@
...
+# Add googletest directly to our build. This defines
+# the gtest and gtest_main targets.
+add_subdirectory(${CMAKE_CURRENT_BINARY_DIR}/googletest-src
+                 ${CMAKE_CURRENT_BINARY_DIR}/googletest-build
+                 EXCLUDE_FROM_ALL)
+
+# The gtest/gtest_main targets carry header search path
+# dependencies automatically when using CMake 2.8.11 or
+# later. Otherwise we have to add them here ourselves.
+if (CMAKE_VERSION VERSION_LESS 2.8.11)
```
anjohan, line 60: the minimum required version is greater than this

CMakeLists.txt
```diff
@@ -0,0 +1,149 @@
...
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND ""
+    INSTALL_COMMAND ""
+)
+ExternalProject_Get_Property(eigen_project SOURCE_DIR)
+add_library(Eigen3 INTERFACE)
+target_include_directories(Eigen3 SYSTEM INTERFACE ${SOURCE_DIR})
+include_directories(${SOURCE_DIR})
+
+# pybind11
```
anjohan, line 81: this looks easier: https://pybind11.readthedocs.io/en/master/compiling.html#building-with-cmake

CMakeModules/FindPythonLibsNew.cmake
```diff
@@ -0,0 +1,195 @@
+# - Find python libraries
```
anjohan, line 1: Is this necessary?

https://cmake.org/cmake/help/latest/module/FindPython.html

CTests/Descriptor/test_bond_env.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    Structure struc1, struc2;
+    LocalEnvironment env1, env2;
+
+    double rcut = 3;
+    std::vector<double> many_body_cutoffs {rcut};
+ 
+    // Prepare cutoff.
+    std::vector<double> cutoff_hyps;
+    void (*cutoff_function)(double *, double, double, std::vector<double>) =
```
anjohan, line 31: @dmclark17 is leaving, you can use `std::function`s now

CTests/Descriptor/test_y_grad.cpp
```diff
@@ -0,0 +1,189 @@
+#include "gtest/gtest.h"
+#include "Descriptor/y_grad.h"
+#include <chrono>
+#include <iostream>
+#include <cmath>
+using namespace std;
```
anjohan, line 6: shame on you

Sources/Descriptor/cutoffs.cpp
```diff
@@ -0,0 +1,47 @@
+#include "cutoffs.h"
+#include <cmath>
+#define Pi 3.14159265358979323846
+
+void quadratic_cutoff(double * rcut_vals, double r, double rcut,
+                      std::vector<double> cutoff_hyps){
```
anjohan, line 6: unused argument

Sources/Environment/local_environment.cpp
```diff
@@ -0,0 +1,354 @@
...
+    Eigen::MatrixXd pos_atom = structure.wrapped_positions.row(atom);
+
+    Eigen::MatrixXd vec1 = structure.cell.row(0);
+    Eigen::MatrixXd vec2 = structure.cell.row(1);
+    Eigen::MatrixXd vec3 = structure.cell.row(2);
+
+    int sweep_no = (2 * sweep_val + 1) * (2 * sweep_val + 1) *
+        (2 * sweep_val + 1);
+
+    double * dists = new double[noa * sweep_no]();
```
anjohan, line 93: these could join the 21st century

Sources/Environment/local_environment.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+    delete [] dists; delete [] xvals; delete [] yvals; delete [] zvals;
+}
+
+void LocalEnvironment :: compute_indices(){
+
+    int no_atoms = rs.size();
+
+    // Initialize a list of lists storing atom indices.
+    int n_cutoffs = n_body_cutoffs.size();
```
anjohan, line 188: can't you just initialize n_body_indices and many_body_indices with an initial size?

Sources/Sparse_GP/sparse_gp.cpp
```diff
@@ -0,0 +1,787 @@
...
+    for (int k = 0; k < n_envs; k ++){
+        for (int i = 0; i < n_sparse; i ++){
+            for (int j = 0; j < n_kernels; j ++){
+                prev_block(i, k) +=
+                    kernels[j] -> env_env(sparse_environments[i], envs[k]);
+            }
+        }
+    }
+
+    // Compute self block. (Note that the work can be cut in half by exploiting the symmetry of the matrix, but this makes parallelization slightly trickier.)
```
anjohan, line 170: let `x` run to `n_envs*(n_envs+1)/2` in parallel, then
```
int k = (-1 + sqrt(1 + 8 * x)) / 2
int l = x - k * (k + 1) // 2
```
(I think, check math)

Sources/Sparse_GP/sparse_gp.cpp
```diff
@@ -0,0 +1,787 @@
...
+    int n_sparse = sparse_environments.size();
+    int n_envs = envs.size();
+    int n_kernels = kernels.size();
+    int n_labels = 3; // Assumes that all three force components are given.
+
+    // Calculate kernels between sparse environments and training environments.
+    Eigen::MatrixXd kernel_block =
+        Eigen::MatrixXd::Zero(n_sparse, n_labels * n_envs);
+
+    #pragma omp parallel for schedule(static)
```
anjohan, line 359: if `n_envs` isn't big, maybe add `collapse(3)`

setup.py
```diff
@@ -1,29 +1,210 @@
...
-)
+
+# Poor man's command-line options parsing
+def steal_cmake_flags(args):
+    """
+    Extracts CMake-related arguments from ``args``. ``args`` is a list of
+    strings usually equal to ``sys.argv``. All arguments of the form
+    ``--cmake-args=...`` are extracted (i.e. removed from ``args``!) and
+    accumulated. If there are no arguments of the specified form,
+    ``NETKET_CMAKE_FLAGS`` environment variable is used instead.
```
anjohan, line 29: maybe remove the NetKet references.

CTests/Descriptor/test_bond_env.cpp
dmclark17: You should use templates to avoid any run-time redirection

CMakeLists.txt
jonpvandermause: There is not. I've removed it

CMakeLists.txt
jonpvandermause: I just followed the wise counsel of Google here; see in particular the section "Incorporating into an existing cmake project":
https://github.com/google/googletest/blob/master/googletest/README.md

anjohan: CMake 3.11+ has FetchContent, which works like externalproject+add_subdirectory.

CMakeLists.txt
jonpvandermause: I removed this if block in a72526e.

CMakeLists.txt
jonpvandermause: Okay, I switched to FetchContent in 68e6443

jonpvandermause: Looks like I'll need to change the minimum required cmake version to 3.14.5:
https://github.com/EMsoft-org/EMsoft/issues/93
(I had trouble compiling with 3.12.1 on Odyssey)

CMakeModules/FindPythonLibsNew.cmake
jonpvandermause: I think it is. If I try find_package(Python ...) instead, I hit a compilation error when making the pybind module triggered by "#include <Python.h>".

If you can find a way to make it work, though, we can switch. I'd certainly prefer to get rid of the dependency on an external CMake module.

YuuuuXie:
> Note: I've included all the flare++ code in this PR, but I've only integrated the structure class with the rest of the repo. In the next week or so, I'll be integrating env and kernels.
> 
> **Sources/Structure**
> 
> * directory containing the C++ implementation of the structure class
> * py_structure.h and py_structure.cpp create the pybind11 binding
> 
> **structure.py**
> 
> * keeps quality-of-life methods (as_dict, str, len, etc.) in pure Python
> 
> **env.py**
> 
> * as a first step toward a C++ implementation of the local environment class, the cutoffs mask has been changed from a dictionary to a class.
> 
> **Sources/Parameters**
> 
> * Bare bones C++ implementation of the hyperparameter mask.
> 
> **setup.py**
> 
> * pip now compiles the C++ code and downloads all dependencies (googletest, Eigen, and pybind11) from github



> Note: I've included all the flare++ code in this PR, but I've only integrated the structure class with the rest of the repo. In the next week or so, I'll be integrating env and kernels.
> 
> **Sources/Structure**
> 
> * directory containing the C++ implementation of the structure class
> * py_structure.h and py_structure.cpp create the pybind11 binding
> 
> **structure.py**
> 
> * keeps quality-of-life methods (as_dict, str, len, etc.) in pure Python
> 
> **env.py**
> 
> * as a first step toward a C++ implementation of the local environment class, the cutoffs mask has been changed from a dictionary to a class.
> 
> **Sources/Parameters**
> 
> * Bare bones C++ implementation of the hyperparameter mask.
> 
> **setup.py**
> 
> * pip now compiles the C++ code and downloads all dependencies (googletest, Eigen, and pybind11) from github

@jonpvandermause Since we have not yet released the 1.0 version, should we merge the c++ code into the dev branch? I'm wondering if it's better to have separate dev1.0 and dev2.0 branches, while this branch can merge into the dev2.0 instead of dev1.0. Just need to make sure dev2.0 pulls dev1.0 all the time

jonpvandermause:
@YuuuuXie It would be a lot of unnecessary work to maintain two development branches. Every time an update to dev1.0 conflicts with dev2.0, the maintainers of dev2.0 would have to deal with a merge conflict, which isn't a sustainable way to develop. I don't see a problem with including the C++ updates in version 1.0, especially since it completely preserves the existing Python interface.

setup.py
jonpvandermause: Removed in 2119a23

Sources/Environment/local_environment.cpp
jonpvandermause: We would first need to count the atoms inside each cutoff sphere to get the size right. Which could certainly be done, and might be slightly faster, but I've instead opted to just push_back the list of indices while we loop over the atoms.

Sources/Descriptor/single_bond.cpp
```diff
@@ -0,0 +1,117 @@
...
+    Eigen::VectorXd &single_bond_vals, Eigen::MatrixXd &force_dervs,
+    Eigen::MatrixXd &stress_dervs,
+    void (*basis_function)(double *, double *, double, int, vector<double>),
+    void (*cutoff_function)(double *, double, double, vector<double>), double x,
+    double y, double z, double r, int s, int environoment_index,
+    int central_index, double rcut, int N, int lmax,
+    const vector<double> &radial_hyps, const vector<double> &cutoff_hyps) {
+
+  // Calculate radial basis values.
+  double *g = new double[N];
```
nw13slx, line 19: for arrays that only exist in a function, please consider using std:vector, instead of dynamically assigned pointer

Sources/Parameters/hyps_mask.h
```diff
@@ -0,0 +1,26 @@
...
+#include <vector>
+
+class HypsMask {
+public:
+  HypsMask();
+
+  // Base attributes on the as_dict method of parameter_helper.py
+  int nspecie = 1, ntwobody = 1, nthreebody = 1, nmanybody = 0, twobody_start,
+      threebody_start, manybody_start, ncut3b = 1;
+  std::vector<int> specie_mask, twobody_mask, threebody_mask, manybody_mask,
```
nw13slx, line 15: I'm not sure I understand how this work. Here's a few observations and thoughts:

- The attributes look slightly different from the python hyps_mask we have. Kernel_name should be an array of strings
- Is there other places that call the hyps_mask content directly by name? In the python one, we can add all different types of mask if there are a new kernel type. As long as the kernel type is declared in Parameters.all_kernel_types, n_kernel_parameters. How do we add new attributes if we have new kernels? 
- We may need to change the default value of ntwobody, ..., nmanybody as -1 to void confusion if we have to hard coded all the attributes in the header.
- if everything has to be hard coded, we should write a test to check the consistency between python interface and the C code...

jonpvandermause: Thanks for the comments! I needed to implement a C++ version of the hyperparameter mask in order to integrate my C++ environment class with the Python version. This is the foundation for all of the many-body kernels that I'm trying to integrate into the main repo ASAP. For that reason, I made the HypsMask class as simple as possible, and I agree that in the future, we should think carefully about how to design a proper class that works with any combination of kernels.

To address your specific concerns:
- The attributes should precisely match the Python hyps_mask, as it exists right now on the tip of the development branch. The kernel names are given in C++ as a vector of strings, which in Python translates to a list of strings.
- The flexibility of the Python version of the mask comes from the fact that it's implemented as a dictionary, not a class, which means that any keyword can be added to it, with no restrictions on the types of the corresponding values. This is flexible but chaotic, and makes it difficult to understand what the mask is doing. We should think about how to implement it in an object-oriented style, while still allowing multiple types of kernels to be accommodated. This should be addressed in the future, but is outside the scope of this PR.
- Instead of assigning an arbitrary default value, it would be better to have ntwobody, nmanybody, etc. directly linked to the corresponding masks, so that when the mask changes, the number changes automatically. This can be done by implementing them as properties of the mask class. We should certainly consider doing this in the future.
- I only hard coded things because that was the fastest way to proceed, but it's certainly not essential. Note that the C++ and Python versions of the mask are consistent enough for all the unit tests to pass on this branch, which should give some confidence, but we can of course consider adding additional tests in the future.

nw13slx: Cool. How about opening a project on Github to list all the features to be implemented in C++? It's easier to keep track of and manage the changes via multiple pull requests.

jonpvandermause: Good idea, I just opened up a project.

Issue #194, opened 2020-06-26
MGP Prediction fails for isolated atoms


stevetorr:
If an atom has no neighbors (such as an isolated gas-phase atom above a surface in vacuum), then the MGP predict method fails when the atomic environment is probed for the nearest 2-body neighbor:

https://github.com/mir-group/flare/blob/development/flare/mgp/mapxb.py#L112

```
        min_dist = atom_env.bond_array_2[0][0]
        lower_bound = np.max(self.maps[0].bounds[0][0])
        if min_dist < lower_bound:
            raise ValueError(f'The minimal distance {min_dist:.3f} is below the'
                f' mgp lower bound {lower_bound:.3f}')
```
If the bond_array_2 list is empty, then an `IndexError` is raised. Ideally 0 forces would be returned for this, but I don't know how this might translate for local energy-- is it simply ill defined for an isolated atom?

Issue #132, opened 2020-01-28
Tutorial (OTF Al) has incomplete lines and unexplained terms.


nw13slx:
Just received complaints from a potential user. The tutorial is a bit vague...

More specifically, the one in docs/source/tutorials/al.rst 
- need examples for two_plus_three body
- need explanation of the hyperparameters, which elements correspond to which variables in the paper
- lines in the code blocks are not complete.

PR #198, opened 2020-07-02
MGP - var upper bound & ASE Atoms

### LINKED ISSUE #22: Tutorials for OTF, MFF, and MFF pairstyle
Before officially releasing the first version of the code, it would be great to include detailed tutorials in the documentation explaining how to use key features (especially otf, mff, and the mff pairstyle in LAMMPS).

YuuuuXie:
MGP:
1. **Added error handler**: when the test env has atomic distance below the lower bound
2. **bugfix**: a pre-factor of 3b energy prediction
3. **improve efficiency**: Reduced the number of 3b energy maps by symmetry, i.e. the map for species `A-A-B` gives same prediction as `A-B-A`, so we only need to map one of them. This reduces number of 3b maps from `N^3` to `N^2(N+1)/2`
4. **interface change**: `mean_only` becomes `var_map`, now we have three options for this arg: `None` (no var map), `pca` (the original implementation of var map using PCA, which is heavy), `simple` (the upper bound var, as lightweight as energy map)
5. **Force mapping removed**

close #22 #132 #194 

ASE:
1. Add a `FLARE_Atoms` class, it is just a wrapper of ASE `Atoms`, meanwhile render the same properties as `Structure`. It can be used in `otf`, so all the `structure` in the ase.calculator can be removed


flare/ase/otf.py
```diff
@@ -83,71 +82,96 @@ class ASE_OTF(OTF):
...
 
     def md_step(self):
-        '''
+        """
         Get new position in molecular dynamics based on the forces predicted by
         FLARE_Calculator or DFT calculator
-        '''
+        """
         # Update previous positions.
         self.structure.prev_positions = np.copy(self.structure.positions)
```
jonpvandermause, line 176: I wonder if there's a way to avoid having to keep around the structure object inside ASE OTF? Maybe after calling the inherited OTF constructor, the structure attribute can be set equal to the atoms object?

flare/kernels/mc_simple.py
```diff
@@ -835,7 +835,7 @@ def three_body_efs_self(env1: AtomicEnvironment, hyps: 'ndarray',
 
 
 def two_body_mc(env1: AtomicEnvironment, env2: AtomicEnvironment,
-                d1: float, d2: float, hyps: 'ndarray', cutoffs: 'ndarray',
+                d1: int, d2: int, hyps: 'ndarray', cutoffs: 'ndarray',
```
jonpvandermause, line 838: Good catch

flare/otf.py
```diff
@@ -384,7 +387,7 @@ def update_positions(self, new_pos: 'ndarray'):
         else:
             self.structure.prev_positions = self.structure.positions
         self.structure.positions = new_pos
-        self.structure.wrap_positions()
+        self.structure.positions[:] = structure.wrap_positions()
```
jonpvandermause, line 390: Good catch that wrap_positions doesn't actually update the wrapped_positions attribute of structure. But I think structure.wrap_positions() should be self.structure.wrap_positions() here? (I'm a bit surprised this didn't cause an error in the OTF unit test)

flare/ase/atoms.py
```diff
@@ -0,0 +1,114 @@
...
+    def coded_species(self):
+        return self.numbers
+
+    @property
+    def forces(self):
+        return self.get_forces()
+
+    @forces.setter
+    def forces(self, forces_array):
+        pass
```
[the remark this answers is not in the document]
YuuuuXie: Yeah I set this function here because in `otf.py` there's the line
```
312         self.structure.forces = forces
```
But we don't need the setter to do anything, because the forces have been written to the `results` dict, so I just put `pass` here

flare/ase/calculator.py
```diff
@@ -40,100 +43,131 @@ def __init__(self, gp_model, mgp_model=None, par=False, use_mapping=False):
...
-            stds[n] = np.sqrt(np.absolute(v))
-            local_energies[n] = e
-
+                f, v, vir, e = self.mgp_model.predict(chemenv)
+                self.results["forces"][n] = f
+                self.results["partial_stresses"][n] = vir
+                self.results["stds"][n] = np.sqrt(np.absolute(v))
+                self.results["local_energies"][n] = e
+
+            except ValueError as err_msg:  # if lower_bound error is raised
```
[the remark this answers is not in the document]
YuuuuXie: Good idea, will re-organize this

flare/ase/dft.py
```diff
@@ -22,9 +22,9 @@ def parse_dft_input(atoms):
 
 
 def run_dft_par(atoms, structure, dft_calc, **dft_kwargs):
```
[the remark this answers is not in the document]
YuuuuXie: Ah, this is just to align the input args with our other dft modules, so in `otf.py` the general interface is
```
305         forces = self.dft_module.run_dft_par(
306             self.dft_input, self.structure, self.dft_loc, n_cpus=self.n_cpus,
307             dft_out=self.dft_output, npool=self.npool, mpi=self.mpi,
308             dft_kwargs=self.dft_kwargs)
```
to make the ase/dft compatible with this interface, I wrote it the way like that

flare/ase/otf.py
YuuuuXie: Right, I actually set the self.structure as the atoms in ase/otf.py
```
135     def get_structure_from_input(self, prev_pos_init):
136         self.structure = self.atoms
```
Yeah this is kind of hard to notice though... The `get_structure_from_input` will be called at the init function, so no `Structure` will appear through the ase otf run

flare/otf.py
YuuuuXie: Oh yeah you're right. I will double check the unit test here

YuuuuXie: I notice in `otf.py` there's the `update_positions` function, which is never used. While in `md.py` there's another `update_positions` which is used in `otf.py`. It seems that they are doing different things. The one in `otf.py` has rescale_temperature. Do you know where to put it? @jonpvandermause

flare/mgp/map3b.py
```diff
@@ -122,98 +120,266 @@ def construct_grids(self):
...
+    else:
+        f0, _ = cutoff_func(r_cut, triplets, 0)  # (n_grid, 3)
+
+    fj = f0[:, 0] * f0[:, 1] * f0[:, 2]  # (n_grid,)
+    fj = np.expand_dims(fj, axis=1)
+
+    return fj, dfj_list
+
+
+def get_hyps_for_kern(hyps, cutoffs, hyps_mask, c2, etypes2):
```
nw13slx, line 183: Maybe it is better to move this part of code to the parameters.py. The point of having a parameter class is to collect all the relevant codes in one file, so it is easier to maintain.

flare/mgp/mapxb.py
```diff
@@ -445,79 +491,86 @@ def update_bounds(self, GP):
...
     def build_map(self, GP):
 
         self.update_bounds(GP)
 
         if not self.load_grid:
             y_mean, y_var = self.GenGrid(GP)
         else:
-            if 'mgp_grids' not in os.listdir(self.load_grid):
-                raise FileNotFoundError("Please set 'load_grid' as the location of mgp_grids folder")
+            if "mgp_grids" not in os.listdir(self.load_grid):
```
nw13slx, line 524: maybe consider adding gp's name or date info to the folder name, so that files in this folder would not be overwritten.

flare/ase/calculator.py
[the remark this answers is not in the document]
YuuuuXie: addressed in commit b2fb7895bb906719302eaaab717786915a29f5aa

flare/otf.py
YuuuuXie: addressed in commit b2fb7895bb906719302eaaab717786915a29f5aa

flare/ase/dft.py
```diff
@@ -36,13 +36,15 @@ def run_dft_par(atoms, structure, dft_calc, **dft_kwargs):
...
+    # The results will be written to the 'results' dictionary, and structure.forces
+    # directly return the atoms.calc.results['forces']
+
+    #    # Write energy, forces, and stress to structure.
+    #    structure.potential_energy = energy
+    #    structure.stress = stress
+    #    structure.stds = np.zeros((structure.nat, 3))
+    #    structure.stress_stds = None
+    #    structure.partial_stress_stds = None
+    #    structure.local_energy_stds = None
```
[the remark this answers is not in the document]
YuuuuXie: addressed in commit fe21344b5ed8011099ef05d8aae1b67e793ba27a

flare/mgp/map3b.py
YuuuuXie: yeah, will move the `get_hyps_for_kern` in map2b and map3b into the parameter.py

flare/mgp/mapxb.py
YuuuuXie: It seems to me that using gp's name or date still can not prevent overwriting, if you just use gp of the same name, or map on the same day. In what situation do you want to keep all the grids files while have the possibility of being overwritten?

PR #201, opened 2020-07-08
[WIP] GPFA Modularization

stevetorr:
In response to user requests and feedback, I have re-written the arguments to `TrajectoryTrainer` to make their meaning more clear. I have also added an extra argument and tiny quality of life tweaks (such as a `__len__` method for GPs).

Given the way the curse of knowledge works (forgetting what it's like to not know), I particularly welcome insight from other users-- particularly those who do not use GPFA much -- to weigh in on if the new arguments are intuitively named and their function is clear.

**Changes**

- **Scripts will break. GPs  are now the first argument to TrajectoryTrainer, supplanting active learning frames as TrajectoryTrainer is meant to facilitate both passive and active learning equally well**.

-  Almost all arguments have been re-named and re-ordered in the `TrajectoryTrainer` module. They are now divided into `passive` and `active` learning arguments to refer to what was previously called the "seed" and "train" parts of the run, confusingly, given that "training" refers to hyperparameter optimization.
 
- `GaussianProcess` now has a `__len__` method which returns the number of atomic environments present.

- New design pattern for GPFA unit tests that would allow the tests to be ran from other directories; we could consider adding this pattern to other files.

### DIFF
--- a/flare/gp_from_aimd.py
+++ b/flare/gp_from_aimd.py
@@ -424,6 +424,7 @@ class TrajectoryTrainer:
 
             if i < train_frame and not self.mgp and len(self.gp) <= \
                     self.max_model_size:
+
                 # Noise hyperparameter & relative std tolerance is not for mgp.
                 if self.mgp:
                     noise = 0
--- a/flare/gp_from_aimd.py
+++ b/flare/gp_from_aimd.py
@@ -79,6 +79,7 @@ class TrajectoryTrainer:
 
                  max_atoms_from_frame: int = np.inf,
                  min_atoms_added_per_train: int = 1,
+                 max_model_size: int = np.inf,
 
                  passive_on_active_skips: int = -1,
                  passive_train_max_iter: int = 50,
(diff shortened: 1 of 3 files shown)


flare/gp_from_aimd.py
```diff
@@ -59,70 +59,91 @@
...
 class TrajectoryTrainer:
 
-    def __init__(self, frames: List[Structure],
-                 gp: Union[GaussianProcess, MappedGaussianProcess],
-                 rel_std_tolerance: float = 4,
-                 abs_std_tolerance: float = 1,
-                 abs_force_tolerance: float = 0,
-                 max_force_error: float = inf,
-                 parallel: bool = False,
+    def __init__(self, gp: Union[GaussianProcess, MappedGaussianProcess],
```
nw13slx, line 62: could you please write something like the backward_attributes and backward_arguments in the GP class to keep track of the change?

stevetorr:
Going to re-factor GPFA in a more substantial way.

Issue #205, opened 2020-07-15
OTF parser fails at "Cutoffs: {'twobody': 3.0, 'threebody': 5.0}"


YuuuuXie:
Currently the otf_parser is hard coded for the header like:
```
Cutoffs: [3.0 5.0]
```
while in the new parameter class, the cutoffs become a dictionary
need to fix the compatibility

YuuuuXie:
Also, in ASE OTF, the cell is written as 
```
Cell([[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]])
```
which is different from the primitive otf output, i.e.
```
[[1.0 0.0 0.0]
 [0.0 1.0 0.0]
 [0.0 0.0 1.0]]
```
need to address this issue as well

YuuuuXie:
The otf_parser.py is not updated, there are so many places not compatible
related to #10

kitpeng11:
same phase also failed in gp model.

YuuuuXie:
close via #214

Issue #49, opened 2019-09-13
OTF restart module/method


YuuuuXie:
Occasionally OTF could be interrupted (either by bug, or the user wants to interrupt the training and change the condition) and needs to restart from middle. It will be good to have a restart module, or a method inside otf module. 
I have a script to do this while I was training my stanene system. If you guys have written any better wrapped module, that will be great.

stevetorr:
I plan to add this as a flag into the gp_from_aimd.py, it should be a simple check to write the model by serialization every few steps.

YuuuuXie:
close via #214

Issue #10, opened 2019-09-04
otf parser tests (test_parse_otf.py) is based on a fixed output file


jonpvandermause:
To test that it works for the current version of otf.py (and its output conventions), the test should be based on a current output file, e.g. the ones generated by test_OTF.py.

YuuuuXie:
close via #214

Issue #212, opened 2020-08-02
How to resume DFT calculation


McAlexDudu:
DFT calculation will get unconverged every now and then, and when this happened, the calculation will stop and we have to restart all over. How to resume? any suggestion?

YuuuuXie:
Currently if you set `write_model` in otf module to 2 or 3, the GP model will be saved automatically, then you can load gp by 
```
gp_model = GaussianProcess.from_file('filename')
```
Also, you can grab the positions of atoms where your calculation was interrupted, save it as an `np.array` of shape `(n_atoms, 3)`, denoted as `previous_pos`.

Then in the otf setup, use
```
my_otf = OTF(prev_pos_init = previous_pos,
             gp = gp_model,
             <other arguments>)
my_otf.curr_step = <your last step number>
my_otf.run()
```

Above is a temporary solution. But we are also considering having a more simple interface for restarting otf, there was an old issue #49 here. We will add it recent days

McAlexDudu:
> Currently if you set `write_model` in otf module to 2 or 3, the GP model will be saved automatically, then you can load gp by
> 
> ```
> gp_model = GaussianProcess.from_file('filename')
> ```
> 
> Also, you can grab the positions of atoms where your calculation was interrupted, save it as an `np.array` of shape `(n_atoms, 3)`, denoted as `previous_pos`.
> 
> Then in the otf setup, use
> 
> ```
> my_otf = OTF(prev_pos_init = previous_pos,
>              gp = gp_model,
>              <other arguments>)
> my_otf.curr_step = <your last step number>
> my_otf.run()
> ```
> 
> Above is a temporary solution. But we are also considering having a more simple interface for restarting otf, there was an old issue #49 here. We will add it recent days

Thanks a lot, I will try it.

PR #214, opened 2020-08-03
Bugfix: OTF parser, ASE md step, OTF checkpoint

### LINKED ISSUE #137: [ASE interface] Logger's compatibility with ASE Trajectory
Currently the data collection is a bit duplicated. 
1. We have a unified *.log file which contains everything about the MD traj and training info, this file is of the same format with primitive GP-OTF log file (without ASE)
2. We also have an `otf_data` folder, containing data files storing GP positions & forces, DFT frames, etc. separately. And this part is duplicated with (1)
3. If ASE's own `trajectory` is on, both the GP and DFT files will be written to it. Consider if we need to separate the two.

One possible solution:
1. For MD data (without DFT), we set an option that either the trajectory can be stored the same as `otf_data` folder, or use ASE trajectory. The only issue for ASE traj is recording the uncertainties, we might need another logger. 
2. Keep DFT data the same as `otf_data` (.xyz files), or write as ASE Trajectory (set an option for the two)

### LINKED ISSUE #158: Structure.from_ase_atoms does not copy forces

YuuuuXie:
### OTF
- [x] close issue: #205 OTF parser fails at cutoffs
- [x] close issue: #10 `test_parser_otf.py` is based on a fixed file
    - added a function in `test_otf_parser` at the end of `test_otf` and `test_ase_otf`, making sure that `OtfAnalysis` and `make_gp` will not report error
    - `test_parser_otf.py` is not yet changed, since it's based on an old file, which is a validation of backward compatibility
- [x] OTF checkpoint: close issue: #49 , #212 

### ASE
- [x] One bug with ASE MD. The ASE's md engine has a function called `step()`, which does something like:
```
def step(self):
    f = atoms.get_forces()
    new_positions = function_update_positions(forces)
    atoms.set_positions(new_positions)
    new_f = atoms.get_forces()
    return new_f
```
and we are using this step function in otf. The consequence is when the atoms' calculator is DFT, both `f` and `new_f` will be calculated by the DFT calculator, but we don't want the `new_f` to be. I changed the `md_step` function in `ase/otf.py` to adapt to this. 

**Note**: when using `NPT` the adaptation is not applied if the `pfactor` is not set to `None`, this will be left as a TODO.

- [x] In `Structure`, the functions `from_ase_atoms` & `to_ase_atoms` now also transfer forces, energy, stress. close #158 
- [x] ASE OTF compatibility with ASE Trajectory. close #137 
    - [x] Add note here, saying that this is not recommended, not tested thoroughly

### Docs
- [x] Update the corresponding docstrings and tutorials


stevetorr:
Just a comment but the body of this PR is really beautifully formatted, nice job! I want to steal some of these ideas (I like the check boxes).

PR #225, opened 2020-08-20
Enable on-the-fly training in the NPT ensemble

jonpvandermause:
ASE OTF runs should now work in the NPT ensemble with nonzero pfactor.

Main changes:
- MD steps are now always taken with the GP energy model to avoid abrupt changes in energies and stresses. This also makes ase.otf.md_step() simpler.
- ASE's NPT class is slightly modified to accommodate the FLARE calculator.


flare/ase/otf.py
```diff
@@ -186,8 +182,9 @@ def compute_properties(self):
             OTF structure object.
         """
 
-        # Reset FLARE calculator if necessary.
+        # Change to FLARE calculator if necessary.
         if not isinstance(self.atoms.calc, FLARE_Calculator):
+            self.flare_calc.results = {}
```
YuuuuXie, line 187: A minor suggestion, maybe add a method in flare calculator.py called `reset()`, i.e.
```
def reset(self):
    self.results = {}
```

Just in case there could be some other parameters resetting needed in the future use

jonpvandermause: Sure, I addressed this in 9cca1c2 and 6f6b0dc

flare/otf.py
```diff
@@ -281,7 +288,6 @@ def run(self):
             # TODO: Reinstate velocity rescaling.
             self.md_step()
             self.curr_step += 1
-            self.checkpoint()
```
YuuuuXie, line 284: This checkpoint was set to solve the user's issue that some OTF training could be interrupted due to external reason, or the interruption during DFT calculation. So the OTF training might not be finished. Then this checkpoint can help user to resume training from that point. Maybe we can set it to be:
```
if write_model == 3:
    self.checkpoint()
```

jonpvandermause: I think lines 294-296 take care of this:
```
        if self.write_model >= 1:
            self.write_gp()
            self.checkpoint()
```
Including an additional call to self.checkpoint seems redundant.

YuuuuXie: That is after the entire OTF is done, then write the state to file. But sometimes the OTF is not complete, but just get interrupted at some step, so maybe better to dump checkpoint each step (or every n steps)

jonpvandermause: Oh right, I see now. I addressed this in fe00a32

PR #229, opened 2020-08-27
Env now has force attribute

stevetorr:
AtomicEnvironments can now get their associated force from the structure they were built from.

Should simplify scripts. I modified one GP function where a force will always be input to the function- making a PR to highlight this change as something most people would find useful in their scripts, as handling forces and environments seperately can get a bit unwieldly.


flare/gp.py
```diff
@@ -1233,14 +1242,13 @@ def from_file(filename: str, format: str = ""):
                         gp_model.alpha = None
                         gp_model.ky_mat_inv = None
                         Warning(
-                            "the covariance matrices are not loaded"
-                            f"it can take extra long time to recompute"
+                            f"the covariance matrices are not loade"
+                            f"dit can take extra long time to recompute"
```
jonpvandermause, line 1246: Looks like there's a minor typo here

PR #230, opened 2020-08-27
Rewrite Active & Passive Learning for GPFA

stevetorr:
Some good suggestions from our users suggest that a re-write of GPFA to support more modular passive and active learning would be desirable.

Some new features for the passive learning module are to allow a maximum model size, cap out the number of elements by element, and preserve some of the previously existing functionality of the `pre_run` function.

- [x] Rewrite passive learning into it's own function, while preserving `pre_run` for now to not break scripts
- [x] Rewrite active learning into a new function, and preserve `run` for now not to break scripts.
- [x] Add a `Trajectory` and `StructureSource` object to `struc.py` as a step towards Flare 2.0 Dev.


flare/gp_from_aimd.py
```diff
@@ -252,18 +252,357 @@ def __init__(
...
+
+            # Atoms which are skipped will have NaN as their force / std values
+            local_energies = None
+
+            # Three different predictions: Either MGP, GP with energy,
+            # or GP without
+            if self.mgp:
+                pred_forces, pred_stds, local_energies = self.pred_func(
+                    structure=cur_frame,
+                    mgp=self.gp,
```
YuuuuXie, line 429: should this be `self.gp` or `self.mgp`?

flare/gp.py
```diff
@@ -657,6 +657,20 @@ def predict(self, x_t: AtomicEnvironment, d: int) -> [float, float]:
...
+        :return:
+        """
+        forces = []
+        stds = []
+        for d in (1, 2, 3):
+            force, std = self.predict(x_t, d)
+            forces.append(force)
+            stds.append(std)
+        return np.array(forces), np.array(stds)
+
```
YuuuuXie, line 673: This function is very useful. Probably change the `gp.predict()` to `gp.predict_force_xyz()` in predict.py, so that the loops are wrapped

flare/gp_from_aimd.py
```diff
@@ -252,18 +252,357 @@ def __init__(
...
+                KE=0,
+                cell=cur_frame.cell,
+            )
+
+            logger.debug(
+                f"Single frame calculation time {time.time()-frame_start_time}"
+            )
+
+            if i < train_frame:
+                # Noise hyperparameter & relative std tolerance is not for mgp.
```
YuuuuXie, line 482: Might be better to wrap the code in this `if` condition into a function (e.g. `add_training_data` or so), so that this main loop looks clearer in terms of code structure

YuuuuXie (changes requested):

flare/gp_from_aimd.py
stevetorr: Ah- `self.mgp` is a flag that indicates if the GP is an `mgp` or not. Should this be renamed to something like `is_mgp`?

YuuuuXie: Ah I see. I thought they represented different modules. Yeah I agree it would be better to rename it like `self.gp_is_mapped` or something similar

flare/gp.py
stevetorr: Great idea. Done!

flare/gp_from_aimd.py
stevetorr: I do like this idea, and appreciate the thoughtful feedback-- I think I will add it in a future refactoring.

PR #241, opened 2020-09-14
Add 2+MB Kernel

stevetorr:
For faster calculations and easier comparison to EAM potentials, add a 2+MB kernel (notably, excluding the three-body contribution).


jonpvandermause (approved):
Looks good, thanks for adding this.

PR #256, opened 2020-10-08
compile lammps+mgp during travis build

jonpvandermause:


.travis.yml
```diff
@@ -11,11 +11,15 @@ before_install:
     - sudo apt update
     - sudo apt install -y quantum-espresso 
     - pip install codecov pytest pytest-cov pytest_mock
+    - git clone https://github.com/lammps/lammps.git lammps
```
anjohan, line 14: consider e.g. `--depth-1` for faster download (although I'm guessing the internet speed is pretty good in the cloud)

jonpvandermause:
Note: on the master branch of lammps, `force.h` doesn't have the `open_potential` method used in `pair_mgp.h`, which causes compilation to fail. This is fixed by cloning the `stable` branch of lammps, but this may become a problem in the future. @YuuuuXie

YuuuuXie:
> Note: on the master branch of lammps, `force.h` doesn't have the `open_potential` method used in `pair_mgp.h`, which causes compilation to fail. This is fixed by cloning the `stable` branch of lammps, but this may become a problem in the future. @YuuuuXie

Got it. It seems they extracted the `open_potential` method into another file `potential_file_reader.h` in the latest version. But we will change to it once they release the new version

PR #278, opened 2021-02-18
Steven/update/stress struc

stevetorr:
## Summary

Stress is now included as an argument to the `Structure` object, so structures can be instantiated with the stress property.
Unit tests have been updated so that serialization / deserialization into JSON includes checking for this field.

- [x] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). 
- [x] Docstrings have been added in the [Sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html).
- [x] Type annotations are **highly** encouraged.
- [x] Tests have been added for any new functionality or bug fixes.
- [x] All existing tests pass.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-numpy>=1.20
+numpy>=1.20.1
 scipy
 memory_profiler
 numba
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-numpy>=1.16.0
+numpy>=1.20
 scipy
 memory_profiler
 numba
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-numpy>=1.16.0
+numpy>=1.20.1
```
jonpvandermause, line 1: Thanks for finding this. Always annoying when dependencies break the CI build

jonpvandermause (approved):

PR #322, opened 2022-10-13
1.3.2 Non normalized dot product kernel

YuuuXie:
* Feature: add `DotProduct` kernel without normalization, to solve the non-smoothness and non-zero energy and forces of atom near the cutoff boundary, when there is only one neighbor in the env (e.g. pulling an isolated H2 molecule). 

**Note:** This issue will only occur when all the neighbors are located right at the cutoff boundary, such that both the numerator (descriptor) and denominator (descriptor norm) will be zero from the cutoff function, and the cancellation of cutoff functions of numerator and denominator result in a non-zero value of d/|d|, but we want zero.

* update version to 1.3.1
* add dot product to lammps (both pair_style and compute), now lammps supports both normalized and non-normalized
* add dot product to python test, especially to the sgp-lammps matching test

Interface change:
* in the lammps coefficient file, the kernel name (`DotProduct` or `NormalizedDotProduct`) is now written and read by lammps. To use coefficient files from older flare version, you can manually add the kernel name in the coeff file


ctests/test_kernels.cpp
```diff
@@ -19,7 +19,7 @@ class KernelTest : public StructureTest {
 
 };
 
-using KernelTypes = ::testing::Types<SquaredExponential, NormalizedDotProduct>;
+using KernelTypes = ::testing::Types<SquaredExponential, NormalizedDotProduct, DotProduct>;
```
cjowen1, line 22: should it be named as DotProduct or NonNormalizedDotProduct?

cjowen1 (approved):
I think everything looks good and consistent. only comment is the naming scheme of the kernel, but I think it probably fine how it is

ctests/test_kernels.cpp
YuuuXie: I think `DotProduct` is fine (and shorter), since it is the "pristine" dot product <di, dj>. While the `NormalizedDotProduct` is the normalized version <di, dj>/(|di|*|dj|)

PR #323, opened 2022-10-13
1.3.3 Relative noise for each training data point

YuuuXie:
### Changes
* Feature: add relative noise for each training data point
    - The actual noise value applied to the kernel of a given data point is `relative_energy_noise * sigma_e` for energy label, `relative_force_noise * sigma_f` for force label, and `relative_stress_noise * sigma_s` for stress label.
    - Currently not supported in offline or online trainings
    - This is to solve the problem that the uniform noise value across different configurations/phases/compounds/sizes does not work well. E.g. the large noise trained from a big complex dataset will bias the prediction on a small simpler dataset
* update version to 1.3.2

### How to use this feature
1. After all the training is done, and a `my_flare.json` file is saved from the training, we load the `json` file in python, and add the `rel_efs_noise` for each training data in the `training_structures` list.
2. Save the modified json into file, and use the loading and mapping methods introduced in the tutorial to get the final model.
3. We suggest to set `relative_energy_noise = n_atoms_per_frame * value` for each frame, such that the total energy noise is proportional to the number of atoms.

### TODOs
- Include the setting of relative noise to on-the-fly or offline training
- Train the per-data relative noise


src/flare_pp/bffs/sparse_gp.cpp
```diff
@@ -1039,10 +1061,12 @@ SparseGP ::compute_likelihood_gradient(const Eigen::VectorXd &hyperparameters) {
   double sigma_f = hyperparameters(hyp_index + 1);
   double sigma_s = hyperparameters(hyp_index + 2);
 
-  Eigen::VectorXd noise_vec = sigma_e * sigma_e * e_noise_one + sigma_f * sigma_f * f_noise_one + sigma_s * sigma_s * s_noise_one; 
-  Eigen::VectorXd e_noise_grad = 2 * sigma_e * e_noise_one;
-  Eigen::VectorXd f_noise_grad = 2 * sigma_f * f_noise_one;
-  Eigen::VectorXd s_noise_grad = 2 * sigma_s * s_noise_one;
+  Eigen::VectorXd noise_vec = sigma_e * sigma_e * inv_e_noise_one 
+                            + sigma_f * sigma_f * inv_f_noise_one
```
cjowen1, line 1065: noise_one variables changed to inv_noise_one. im sure you have confirmed that this is but the inconsistency (with previous naming) is why I am commenting

cjowen1 (approved):
all looks consistent except for one comment that I made.

src/flare_pp/bffs/sparse_gp.cpp
YuuuXie: yes, so basically I have kept both `e_noise_one` and `inv_e_noise_one`. Previously without relative noise, this value (the relative noise) is just 1. And now I use those two to save `rel_e_noise` and `1/rel_e_noise`.

PR #324, opened 2022-10-21
add checks for ase calculators with no stress tensor property impleme…

juliayang:
## Summary

* Add fixes for when stress is still calculated despite user indication that `stress_training = False`
* Import PropertyNotImplementedError from ase for readability


YuuuXie:
Merge to dev branch and update the docs for continuous integration

PR #328, opened 2022-10-28
1.3.1 Bug fixings - stress, pair_style cutsq, offline preprocess, sanity check

YuuuXie:
* Fix atoms.py and otf.py when stress is not available in dft (based on Julia's fixing)
* Add `cutsq` to pair_flare.cpp and compute_flare_std_atom.cpp, such that `cutsq` is the same as the `cutoff_matrix`. And it is used internally by lammps, when using `fix swap/atom`
* Preprocessing the trajectory file for offline training (FakeMD), such that the original "step" saved from otf training in the xyz file is removed (fake.py)
* Add a sanity check for the shape of `cutoff_matrix` in the python script when reading yaml, since flare will still run without error if given a wrong-shaped cutoff_matrix, and will give zero forces

- [x] update version number to 1.3.1


flare/scripts/otf_train.py
```diff
@@ -229,6 +229,10 @@ def get_sgp_calc(flare_config):
     cutoff = flare_config["cutoff"]
     descriptors = []
     for d in flare_config["descriptors"]:
+        if "cutoff_matrix" in d:  # multiple cutoffs
+            assert np.allclose(np.array(d["cutoff_matrix"]), (n_species, n_species)),\
```
anjohan, line 233: Is this not missing `.shape`?

anjohan (approved):

PR #344, opened 2023-01-23
Fixed train_hyps accepting tuple with "inf".

ProodjePindakaas:
Setting train_hyps = (i,"inf") would give an error as the tuple cannot be assigned new upper bound np.inf. Also changes bounds of dft_count to not include upper bound, as this seems more intuitive. Documentation in OTF class changed accordingly.

Adapted from the Pymatgen PR template.

## Summary

Include a summary of major changes in bullet points:

* Feature 1
* Feature 2
* Fix 1
* Fix 2

## Additional dependencies introduced (if any)

* List all new dependencies needed and justify why. While adding dependencies that bring
significantly useful functionality is perfectly fine, adding ones that 
add trivial functionality, e.g., to use one single easily implementable
function, is frowned upon. Provide a justification why that dependency is needed.

## TODO (if any)

If this is a work-in-progress, write something about what else needs 
to be done

* Feature 1 supports A, but not B.

## Checklist

Work-in-progress pull requests are encouraged, but please put [WIP]
in the pull request title.

Before a pull request can be merged, the following items must be checked:

- [ ] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/). 
- [ ] Docstrings have been added in the [Sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html).
- [ ] Type annotations are **highly** encouraged.
- [ ] Tests have been added for any new functionality or bug fixes.
- [ ] All existing tests pass.
- [ ] The version number in `flare/_version.py` is updated. We are using a version number format a.b.c
    - If this PR fixes bugs, update version number to a.b.c+1
    - If this PR adds new features, update version number to a.b+1.0
    - If this PR includes significant changes in framework or interface, update version number to a+1.0.0

### DIFF
--- a/flare/learners/otf.py
+++ b/flare/learners/otf.py
@@ -95,8 +95,10 @@ class OTF:
             DFT is called. Defaults to 1.
         train_hyps (tuple, optional): Specifies the range of steps the
             hyperparameters of the GP are optimized. If the number of DFT
-            calls is in this range, the hyperparameters are frozen.
-            Defaults to (None, None) which means always training.
+            calls is ouside this range, the hyperparameters are kept fixed.
+            To always optimize hyperparameters use (0,"inf").
+            Defaults to (0, 1) which means it only optimizes on the first dft 
+            call.
         min_steps_with_model (int, optional): Minimum number of steps the
             model takes in between calls to DFT. Defaults to 0.
         dft_kwargs ([type], optional): Additional arguments which are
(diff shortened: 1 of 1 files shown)


YuuuXie (approved):
Hi @ProodjePindakaas, thanks for the pull request. I don't think it is a problem since you could just use a list `[0, inf]` instead of a tuple, and then the 2nd parameter can be changed. Nevertheless, thanks for pointing out the inaccurate descriptions of the docstring. We are going to make this `train_hyps` more flexible than its current form in a next version.

PR #351, opened 2023-04-27
bug fixes: Kuu_jitter, sparse_descriptor initialization

YuuuXie:
Bug fixes: 

1. Kuu_jitter was not saved
2. The sparse descriptors initialization could be duplicated so its length is 1 longer than it is supposed to


cjowen1 (approved):
looks good

PR #414, opened 2024-09-16
relax numpy requirement

jonpvandermause:
## Summary

* removes unspecified np.ints and np.floats (deprecated in recent versions of numpy)
* relaxes numpy requirement in requirements.txt


YuuuXie (approved):
Looks good, thanks!

